PR #1, opened 2019-02-27
Aggregated gene level output

jafors:


johanneskoester:
Nice work. Can you pull latest master from here? We have travis enabled now, and I want to see if the tests pass.

PR #6, opened 2019-10-22
r-base to 3.6, default off (spia, fgsea, goatools), fix sleuth RAM problem

dlaehnemann:
* `r-base = 3.6` in all environments
* introduce `activate: false` defaults for spia, fgsea and goatools (set to `true` to run the tools)
* restrict thread usage of `sleuth_prep()` in rule `sleuth_init` to whatever the pipeline provides (with a current default max at `6`), instead of letting it take whatever it can find -- this can lead to high RAM usage and running R code in the form of scripts seems to have a hard limit at 40 GB RAM

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -54,51 +54,46 @@ def get_trimmed(wildcards):
     return expand("results/trimmed/{sample}-{unit}.fastq.gz", **wildcards)
 
 def is_activated(config_element):
-    return config_element.get('activated', 'false') in {"true","True"}
-  
+    return config_element['activate'] in {"true","True"}
+
 def get_bootstrap_plots(model, gene_list=None):
     """Dynamically determine which transcripts to plot based on
        checkpoint output."""
-    def inner(wildcards):
-        transcripts = dict()
-        genes = set()
-        # Obtain results from the sleuth_diffexp checkpoint.
-        # This happens dynamically after the checkpoint is completed, and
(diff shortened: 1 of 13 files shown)


dlaehnemann:
@johanneskoester this passes, a review of the changes would be very much appreciated.

workflow/rules/common.smk
```diff
@@ -53,49 +53,47 @@ def get_trimmed(wildcards):
...
+        genes.add("Custom")
+    for g in genes:
+        if not pd.isnull(g):
+            valid = results.ext_gene == g
+            trx = set(results[valid].target_id)
+            transcripts[g] = trx
+    # Require the respective output from the plot_bootstrap rule.
+    return ["results/plots/bootstrap/{gene}/{gene}.{transcript}.{model}.bootstrap.pdf".format(gene=g, transcript=t, model=model)
+            for g, ts in transcripts.items()
+            for t in ts]
```
johanneskoester: No inner anymore?

dlaehnemann: Nope, this is now called in an input function to the `rule all` in the `Snakefile`, but in a `for` loop without list comprehension. Not as elegant as a `list comprehension`, but I find it more readable. But have a look here:

https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/pull/6/files#diff-54ba7003a36532ba119a71460aca1b42R87-R90

PR #7, opened 2019-10-30
Added title to heatmap and mention in report annotation that the heat…

HenningTimm:
…map shows tpm values.


workflow/rules/common.smk
```diff
@@ -35,7 +35,16 @@ wildcard_constraints:
 
 def is_single_end(sample, unit):
     """Determine whether unit is single-end."""
-    return pd.isnull(units.loc[(sample, unit), "fq2"])
+    fq2_present = pd.isnull(units.loc[(sample, unit), "fq2"])
+    if isinstance(fq2_present, pd.core.series.Series):
+        # if this is the case, get_fastqs cannot work properly
+        raise ValueError(
```
HenningTimm: This adds an error message for a common error case. When there is a duplicate entry in the units.tsv, i.e. a unit name is used twice for a given sample
```
Sample    unit     fq1               fq2
A              u1      foo_1.fq       foo_2.fq
A              u1      bar_1.fq       bar_2.fq    <<< this line has a duplicate unit name
B              u1      baz_1.fq      baz_2.fq
```
the result of line 38 is a `Series`, not a boolen. Until now this prompted an error due to ambiguous truth values (cf. [this](https://stackoverflow.com/q/36921951/2862719)). I did not find other cases that result in the same error, so not raising the actual error (truth value ambiguous) seems safe to me.

dlaehnemann: Then, a follow-up question: Is the first sentence of the error message correct? If I in turn understand the problem correctly, then having the same sample-unit combination twice means that `units.loc` does not return a single `fq2` value, but a series. Thus the issue really is just the duplicate usage of the same sample-unit combination in the `units.tsv` and that is all that the error message needs to say? Or am I missing something?

HenningTimm: I agree that the first sentence is not a helpful error message. It describes where in the code the problem occurs, which was helpful to me but most certainly is not helpful for another user. I wanted to avoid an overly specific error message that could mask other errors, since I am not sure if other errors in the file can cause the same Exception.

Maybe replacing the first sentence with something like "Could not unambiguously identify units.\n" would be better?

dlaehnemann: Maybe "Multiple fq2 entries found for sample-unit combination {sample}-{unit}.\n"? Would you just change this and push the change to this PR? I could then approve and merge it.

HenningTimm: Done :)

dlaehnemann:
So much for merging, I don't have write access... Will let @johanneskoester know...

johanneskoester:
@HenningTimm can you update to latest master?

HenningTimm:
@johanneskoester Done. I had a merge conflict with the ngs-testdata submodule, which is strange, since I did not do any work there.

HenningTimm:
@johanneskoester This was a problem with my local version of the workflow and can be considered resolved. @dlaehnemann and I did review this and it should be good to merge.

PR #12, opened 2020-01-29
volcano plot script and respective rules

dlaehnemann:
This is a pull request to enable easier comparisons to the master branch and for enabling inline code review.

Pinging in @AntonieV , @TharjaX for notifications about this new pull request.

### DIFF
--- a/workflow/scripts/plot-volcano.R
+++ b/workflow/scripts/plot-volcano.R
@@ -1,6 +1,6 @@
-log <- file(snakemake@log[[1]], open="wt")
-sink(log)
-sink(log, type="message")
+#log <- file(snakemake@log[[1]], open="wt")
+#sink(log)
+#sink(log, type="message")
 
 #---install required packages for local testing without surrounding snakemake-workflow---
 #install.packages("BiocManager")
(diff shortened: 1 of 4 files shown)


workflow/scripts/plot-volcano.R
```diff
@@ -0,0 +1,233 @@
...
+
+print("Building volcano plot...")
+
+
+
+#setwd("/home/tharja/Schreibtisch/RNA-Seq-Project/tmp_rna-seq-kallisto-sleuth/")
+ 
+pval_transcripts <- sleuth_load(snakemake@input[["pvals"]])
+conditions <- read.table(snakemake@input[["samples"]])
+matrix <- read.table(snakemake@input[["matrix"]], quote="\"", check.names = F, stringsAsFactors=F, header = TRUE)
```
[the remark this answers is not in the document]
dlaehnemann: Although we might want to explore using the default sleuth function `plot_volcano` first (see comment below at line 150), which might work with only reading in the `pvals` object.

workflow/scripts/plot-volcano.R
```diff
@@ -0,0 +1,233 @@
...
+  #example <- c(1, 2, 3, 7, 4, 9, 5)
+  #plot(example)
+  #dev.off()
+  #--------------------------------------------
+  write_tsv(fold_change, snakemake@output[["foldchange"]])
+  pdf(file = snakemake@output[["volcanoplot"]], width = 14)
+  #par()
+
+  #pdf(file = "results/plots/volcano/model_X.volcano.png", width = 14)
+    volcano <- ggplot(plot_data, aes(x=plot_data$log2FC, y=plot_data$pval))+
```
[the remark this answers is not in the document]
dlaehnemann: Hmm, I suspect the log2-fold change instead of the default beta values might be the reason for re-implementing?

workflow/scripts/plot-volcano.R
```diff
@@ -0,0 +1,233 @@
...
+
+  # plot_data <- plot_data[is.finite(plot_data$pval)]
+ plot_data$pval[!(is.finite(plot_data$pval))] <- 0
+ plot_data$pval[is.na(plot_data$pval)] <- 0
+ plot_data$pval[is.nan(plot_data$pval)] <- 0
+
+  # plot_data <- plot_data[is.finite(plot_data$qval)]
+ plot_data$qval[!(is.finite(plot_data$qval))] <- 0
+ plot_data$qval[is.na(plot_data$qval)] <- 0
+ plot_data$qval[is.nan(plot_data$qval)] <- 0
```
dlaehnemann: I think we might also achieve this more elegantly with piping `plot_data` through some [`drop_na()`](https://tidyr.tidyverse.org/reference/drop_na.html) and some [`filter()`](https://dplyr.tidyverse.org/reference/filter.html) instances for the `Inf` values. Although we should also double-check, where these values come from in the first place, to make sure it's not a bug elsewhere. This is probably easiest done looking at actual data from an example run together.

workflow/scripts/plot-volcano.R
```diff
@@ -0,0 +1,233 @@
...
+   # print(plot_data$pval)
+
+  #---plot example for testing snakemakerule---
+  #pdf(file = snakemake@output[[1]], width = 14)
+  #example <- c(1, 2, 3, 7, 4, 9, 5)
+  #plot(example)
+  #dev.off()
+  #--------------------------------------------
+  write_tsv(fold_change, snakemake@output[["foldchange"]])
+  pdf(file = snakemake@output[["volcanoplot"]], width = 14)
```
dlaehnemann: Instead of the old `pdf()` + `dev.off()` R syntax, I would opt for the ggplot2 function `ggsave()` run after creating the plot. See below, for the spot to put it.

workflow/scripts/plot-volcano.R
```diff
@@ -0,0 +1,233 @@
...
+  #---plot example for testing snakemakerule---
+  #pdf(file = snakemake@output[[1]], width = 14)
+  #example <- c(1, 2, 3, 7, 4, 9, 5)
+  #plot(example)
+  #dev.off()
+  #--------------------------------------------
+  # ggsave(filename=snakemake@output[[1]], width = 14)
+# #  write_tsv("foldchange", snakemake@output[["foldchange"]])
+   # print(volcano)
+  dev.off()
```
dlaehnemann: I would opt for the more modern [`ggsave()`](https://ggplot2.tidyverse.org/reference/ggsave.html) function from the comments, here.

dlaehnemann:
I'll close this in favour of PR #14, where the plot implementation is currently happening.

PR #13, opened 2020-02-04
any name for diffexp grouping column, fix no custom_transcripts, future-proof column select()

dlaehnemann:
The main point of this PR is to remove the schema-requirement of a sample.tsv sheet column named `condition`. Usually, it makes more sense to specify a more informative name for that column (e.g. something like `group1_vs_group2`), which will also be picked up in labeling etc. This also requires removing the `condition` column from the respective `select()`, but the custom-named column will be picked up through the formula, anyways.

As a side-effect, this PR future-proofs the column `select()` on a vector of strings with the `all_of()` function and fixes a bug that occurs, if the provided transcriptome does NOT contain any custom transcripts -- i.e. we now handle this case properly.

### DIFF
--- a/workflow/scripts/sleuth-init.R
+++ b/workflow/scripts/sleuth-init.R
@@ -37,7 +37,7 @@ if(!is.null(snakemake@params[["exclude"]])) {
 
 if(!is.null(model)) {
     # retrieve the model formula
-    formula <- as.formula(snakemake@params[["model"]])
+    formula <- as.formula(model)
     # extract variables from the formula and unnest any nested variables
     variables <- labels(terms(formula)) %>%
                     strsplit('[:*]') %>%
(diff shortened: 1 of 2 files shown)


workflow/scripts/sleuth-init.R
```diff
@@ -37,18 +37,19 @@ if(!is.null(snakemake@params[["exclude"]])) {
...
+    formula <- as.formula(model)
     # extract variables from the formula and unnest any nested variables
     variables <- labels(terms(formula)) %>%
                     strsplit('[:*]') %>%
                     unlist()
     # select the columns required by sleuth and filter to all samples where
     # none of the given variables are NA
     samples <- samples %>%
-                select(sample, path, condition, variables) %>%
+                select(sample, path, all_of(variables)) %>%
```
johanneskoester: Can you please ensure that afterwards all columns that are not part of the formula are still in the sample sheet? We just want to remove rows here, not columns. The reason is that people might want to use certain columns e.g. in the bootstrap or PCA plots, even if they are not part of the formula.

dlaehnemann: This is now in PR #19

PR #14, opened 2020-02-07
Plots v2

AntonieV:
Implementation of volcano plot:
For getting sleuth object after wald-test directly, sleuth::plot_volcano() is called in sleuth-diffexp.R script 
without an own snakemake-rule. For this plot it's necessary to set value of pval_aggregate of sleuth object to FALSE. Set pval_aggregate after that to TRUE again, leads to an unexpected behavior in sleuth and error message, which is reported here:
https://github.com/pachterlab/sleuth/pull/240


workflow/scripts/plot-scatter.R
```diff
@@ -0,0 +1,30 @@
...
+               units = "est_counts", offset = 1, point_alpha = 0.2, xy_line = TRUE,
+               xy_line_color = "red", trans = "log", xlim = NULL, ylim = NULL)+ 
+    labs(x = sample1, y = sample2)
+}
+
+all_plots <- ggarrange(plotlist = plot_list)
+plot_figure <- annotate_figure(all_plots, 
+                top = text_grob("log transformed scatter plots of different samples",
+                                color = "black", face = "bold", size = 14)
+                )
```
dlaehnemann: Here, I am not sure what we want. What do you think?

I see two options:

1. As you implemented it. This way, each combination is only plotted once and we avoid redundant plots.
2. Plot every combination of samples in a grid, similar to the arrangement [e.g. seen here](http://www.sthda.com/english/wiki/scatter-plot-matrices-r-base-graphs). This way, each combination is easier to find at a glance. To generate this, you could just iterate over all samples in two nested loops, i.e.:

```
for (i in so$sample_to_covariates$sample) {
  for (j in so$sample_to_covariates$sample) {
    add plot_scatter to list
 }
}
```

And then wrap the plots after every `length(so$sample_to_covariates$sample)`.

So, what do you think?

workflow/report/loadings-plot.rst
```diff
@@ -1 +1,2 @@
-Shows principal components of {{ snakemake.wildcards.covariate }}, which transcripts contribute the most.
+**Plot loadings**, computed with sleuth, shows for each principal components of {{ snakemake.wildcards.covariate }}, which transcripts (estimated counts) contribute the most.
```
dlaehnemann: Not sure if this is needed, but for consistency I would add the leading and trailing double \`\` before and after the jinja template string (i.e. before and after the `{{}}`)

workflow/report/plot-ma.rst
```diff
@@ -1 +1,3 @@
-MA plots for wald test and with significance level of {{ snakemake.params.sig_level_ma }}. The plots display, for each transcript, the mean of abundances across samples on the x-axis and fold change on the y-axis.
+**MA plots** computed with sleuth for wald test using the model ``{{ snakemake.params.model["full"] }}`` for ``{{ snakemake.wildcards.covariate }}``.
+The plots display, for each transcript, the mean of abundances across samples on the x-axis and fold change on the y-axis.
+Significant genes up to significance level of {{ snakemake.params.sig_level_ma }} are coloured red.
```
dlaehnemann: Also missing the leading and trailing \`\`

workflow/report/pc-variance-plot.rst
```diff
@@ -0,0 +1 @@
+Plot that compares specified principal compontent variances retained by percentage.
```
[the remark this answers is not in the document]
dlaehnemann: Thanks for all the detailed descriptions! Just some minor things I further commented on.

workflow/scripts/plot-scatter.R
dlaehnemann: I'm not too worried about nested for loops with a limited number of samples---but who knows, maybe people will throw in excessive amounts of samples in the future. So you may be right in being more systematic, here. So let's merge this plot as you implemented it now and have a look on some real data.

dlaehnemann:
@johanneskoester I don't have the necessary permissions to merge this, so could you? This should all be fine and reviewed---and we can add further adjustments in a new branch. But I would prefer to have this merged for now, to pull this into an actual workflow and test there --- and so that we don't end up with a diverging master...

workflow/rules/diffexp.smk
```diff
@@ -142,6 +159,31 @@ rule tpm_matrix:
     script:
         "../scripts/sleuth-to-matrix.R"
 
+rule plot_group_density:
+    input:
+        "results/sleuth/all.rds"
+    output:
+        report("results/plots/group_density/{model}.group_density.pdf", caption="../report/group-density.rst", category="Densities")
```
dlaehnemann: One last thing, sorry...

I would probably just stick those plots into the QC category for now, otherwise the report might get a bit unwieldy. If in the end this category gets overfilled, we'll have to think of a strategy for more meaningful categories.

workflow/rules/diffexp.smk
```diff
@@ -142,6 +159,31 @@ rule tpm_matrix:
...
+    log:
+        "logs/plots/group_density/{model}.group_density.log"
+    script:
+        "../scripts/plot-group-density.R"
+
+rule plot_scatter:
+     input:
+         "results/sleuth/all.rds"
+     output:
+         report("results/plots/scatter/{model}.scatter.pdf", caption="../report/scatter.rst", category="Scatter")
```
dlaehnemann: Same as above, I would just stick this in the category QC for now.

.test/config/config.yaml
```diff
@@ -31,6 +35,10 @@ diffexp:
       # Binary valued covariate that shall be used for fold change/effect size
       # based downstream analyses.
       primary_variable: condition
+  # significance level used for volcano, ma- and qq-plots
+  sig_level_volcano-plot: 0.05
```
johanneskoester: have a section `sig-level`, and below entries for qq, ma, and volcano plots.

dlaehnemann: Good idea!

PR #16, opened 2020-02-14
merge samples in kallisto_quant step

dlaehnemann:
This PR implements sample merging by providing kallisto with all the units of a sample at once. I.e. if multiple sequencing runs are provided for a sample as units, kallisto will jointly quantify abundance for all the runs. In addition, the workflow now checks whether units of a sample are either all paired end or all single end and throws an exception if this is violated for a sample.


johanneskoester:
The implementation is fine. I am just wondering if this is really what one wants to do. If units are just unmerged lanes, then yes. If they are technical replicates, then probably no. Let us have a chat about this when I am in the office on Thursday.

dlaehnemann:
Sure, let's discuss further. My first instinct would be for technical replicates to be modelled via the batch variables...

dlaehnemann:
In practice, I have only ever seen multiple units per sample when we have multiple lanes that a sample was sequenced across. Sometimes, because a biological sample was re-sequenced because of low coverage, sometimes systematically (better library mixing, I guess?). And whenever we have any type of replication in the experiment, we do model this as a batch effect.

Thus, I would argue for this merging of units, again. We'll need a new PR off of the current code, but can probably copy over most stuff. What do you think @johanneskoester?

PR #21, opened 2020-03-31
Bootstrap plots checkpoint to report directory

dlaehnemann:
This PR also harmonizes the naming of the rst-files for plots used in the HTML report.

Also, it seems like the wildcards used in the new `directory(patterns=)` keyword are not available upon parsing of the rst file (as is suggested [in the snakemake docs](https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html), at least the `snakemake.wildcards.transcript` from the `rule plot_bootstrap` simply does not appear in the caption in the produced report.


dlaehnemann:
@johanneskoester This requires the GitHub Actions Docker to be updated to `snakemake 5.13.0` via this PR (plus then a new release of GitHub Actions): https://github.com/snakemake/snakemake-github-action/pull/5

johanneskoester:
Done!

dlaehnemann:
@johanneskoester Then feel free to review and merge... ;)

dlaehnemann:
Just realised one more question: Should we attempt the same with the fgsea checkpoint? It should also work and would simply produce several plots in one run of the rule instead of only one plot per run of the rule...

johanneskoester:
> This PR also harmonizes the naming of the rst-files for plots used in the HTML report.
> 
> Also, it seems like the wildcards used in the new `directory(patterns=)` keyword are not available upon parsing of the rst file (as is suggested [in the snakemake docs](https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html), at least the `snakemake.wildcards.transcript` from the `rule plot_bootstrap` simply does not appear in the caption in the produced report.

This should work now with snakemake 5.14.

johanneskoester:
Given that it works now, do you want to change the caption text to make use of it?

dlaehnemann:
Oh, the captions were using it anyways -- it didn't throw an error and just kept them blank, so I left that in awaiting the fix. But I now touched up the captions a bit more, adding in the respective wildcard for the fgsea gene set plots and putting separate sentences on separate lines.

PR #22, opened 2020-04-07
Ihw fdr control

AntonieV:
Independent Hypothesis Weighting (IHW) for false discovery rate control: integration in workflow, implementation of script with an additional error handling for grouping before ihw calculation. Rule and reports added.


workflow/scripts/ihw-fdr-control.R
```diff
@@ -0,0 +1,116 @@
...
+
+#select the necessary data
+if(level_name == "genes-aggregated") {
+  gene_data <- gene_data %>%
+    select(-c(qval, sum_mean_obs_counts, num_aggregated_transcripts)) %>%
+    mutate(grouping = groups_by_filter(mean_obs, number_of_groups))
+} else {
+  gene_data <- gene_data %>%
+    select(ens_gene, ext_gene, pval, mean_obs) %>%
+    mutate(grouping = groups_by_filter(mean_obs, number_of_groups))
```
dlaehnemann, line 52: I think here I would add another column with the grouping as a factor, for the plotting below, i.e.:
```
    mutate(
      grouping = groups_by_filter(mean_obs, number_of_groups),
      grouping_factor = factor(grouping, ordered = TRUE)
    )
```

workflow/scripts/ihw-fdr-control.R
```diff
@@ -0,0 +1,116 @@
...
+
+ggsave(snakemake@output[["dispersion"]], dispersion)
+
+#histograms
+hist_overall <- ggplot(gene_data, aes(x = pval)) +
+  geom_histogram(binwidth = 0.025, boundary = 0) +
+  xlab("p-values without grouping") +
+  ylab("density")
+
+levels(gene_data$grouping) <- paste0(rep("group ", number_of_groups), 1:number_of_groups)
```
dlaehnemann, line 72: I think you can skip this line if the required factor is introduced in the `mutate` on `gene_data` I suggest above.

workflow/scripts/ihw-fdr-control.R
```diff
@@ -0,0 +1,116 @@
...
+hist_groups <- ggplot(gene_data, aes(x = pval)) +
+  geom_histogram(binwidth = 0.025, boundary = 0) +
+  xlab("p-values of the individual groups") +
+  ylab("density") +
+  facet_wrap(~ grouping, nrow = number_of_groups %/% 4 + ifelse((number_of_groups %% 4 != 0), 1, 0))
```
dlaehnemann, line 77: Here, I would use the newly created `grouping_factor` I suggest above. Also, I think you can get to the intended value for `nrow` a bit easier with the base R function `ceiling`:
```
  facet_wrap(~ grouping_factor, nrow = ceiling(number_of_groups / 4) )
```

workflow/scripts/ihw-fdr-control.R
```diff
@@ -0,0 +1,116 @@
...
+
+ggexport(histograms,
+         filename = snakemake@output[["histograms"]],
+         width=14)
+
+# ihw calculation
+write_tsv(gene_data, snakemake@output[["results"]])
+ihw_results_mean <- ihw(pval ~ mean_obs, data = gene_data, alpha = 0.1, nbins = tested_number_of_groups)
+ihw_mean <- as.data.frame(ihw_results_mean)
+write_tsv(ihw_mean, snakemake@output[["results"]])
```
dlaehnemann, line 95: The second `write_tsv()` writes to the same file as the first, so it will simply overwrite its contents. Does that mean we can simply drop the first `write_tsv()` statement? Is that an artefact of script development? :D

dlaehnemann:
@johanneskoester Could you review and possibly merge this? When merging, could you also mention the important new side-effect of this PR in the commit message:

* [] `sleuth_init.R` will now use "useast" instead of "www" as the default ensembl biomaRt mirror and cycle through all 4 available mirrors (up to three times) to find one that does not throw an error

workflow/rules/enrichment.smk
```diff
@@ -10,7 +10,9 @@ rule download_bioconductor_species_database:
     log:
         "logs/resources/bioconductor/{package}.log"
     shell:
-        "conda create --yes --quiet -p {params.path} --channel bioconda bioconductor-{wildcards.package}={params.version}"
+        "conda clean --all --yes --quiet; "
```
johanneskoester, line 13: Conda clean --all might destroy environments according to the docs. Although I have never seen this, we cannot enforce this for the end user. May I ask you to try this command before the workflow run? The snakemake-github-action has a `stagein` parameter for this purpose. Perhaps something is leaked in from the dockerfile.

workflow/scripts/sleuth-init.R
```diff
@@ -8,11 +8,53 @@ library("tidyverse")
...
+  mart <- tryCatch(
+    {
+      # done here, because error function does not
+      # modify outer scope variables, I tried
+      if (mart == "www") rounds <- rounds + 1
+      # equivalent to useMart, but you can choose
+      # the mirror instead of specifying a host
+      biomaRt::useEnsembl(
+        biomart = "ENSEMBL_MART_ENSEMBL",
+        dataset = "hsapiens_gene_ensembl",
```
dlaehnemann, line 26: Adjust to flexible species selection again!

AntonieV:
@johanneskoester thanks :)

PR #23, opened 2020-04-09
Differential splicing analysis.

johanneskoester:
This PR introduces differential splicing analysis with IsoformSwitchAnalyzeR.


.test/config/config.yaml
```diff
@@ -43,6 +43,13 @@ diffexp:
     ma-plot: 0.05
     qq-plot: 0.05
 
+diffsplice:
+  # codingCutoff parameter of isoformSwitchAnalyzer, see
+  # https://rdrr.io/bioc/IsoformSwitchAnalyzeR/man/analyzeCPAT.html
+  coding_cutoff: 0.725
+  # Should be set to true when using de-novo assembled transcripts.
+  remove_noncoding_orfs: false
```
johanneskoester, line 51: Need to add the alpha/sig-level here. I would like to harmonize the notation for significance in the pipeline. Maybe even just one global threshold? Allowing to set it differently everywhere seems both error prone and even problematic from a philosophic perspective.

workflow/rules/refs.smk
```diff
@@ -0,0 +1,82 @@
...
+        ncrna="results/refs/transcriptome.ncrna.fasta"
+    output:
+        "results/refs/cpat.logit.RData"
+    params:
+        prefix=lambda _, output: output[0][:-12]
+    conda:
+        "../envs/cpat.yaml"
+    shell:
+        "make_logitModel.py --hex={input.hexamers} --cgene={input.cds} "
+        "--ngene={input.ncrna} -o {params.prefix}"
```
johanneskoester, line 81: @dlaehnemann any other rules that retrieve reference data should be moved in here as well. Feel free to hijack this PR for that purpose.

workflow/rules/refs.smk
```diff
@@ -0,0 +1,82 @@
...
+        "make_hexamer_tab.py --cod={input.cds} --noncod={input.ncrna} > {output}"
+
+
+rule calculate_cpat_logit_model:
+    input:
+        hexamers="results/refs/cpat.hexamers.tsv",
+        cds="results/refs/transcriptome.cds.fasta",
+        ncrna="results/refs/transcriptome.ncrna.fasta"
+    output:
+        "results/refs/cpat.logit.RData"
```
johanneskoester, line 74: The paths have to be adjusted though. This should go into "resources/...".

workflow/scripts/isoform-switch-analysis-init.R
```diff
@@ -0,0 +1,51 @@
+log <- file(snakemake@log[[1]], open="wt")
+sink(log)
+sink(log, type="message")
+
+library("tidyverse")
+library("sleuth")
```
dlaehnemann, line 6: I can't see anything below that actually uses any sleuth functionality. Can we drop it here and from the environment `yaml` file?

workflow/envs/isoform-switch-analyzer.yaml
```diff
@@ -0,0 +1,8 @@
+channels:
+  - bioconda
+  - conda-forge
+dependencies:
+  - r-sleuth =0.30
```
dlaehnemann, line 5: I don't think, sleuth is needed in this environment.

workflow/rules/diffsplice.smk
```diff
@@ -0,0 +1,92 @@
...
+    output:
+        rds="results/diffsplice/{model}.diffsplice.rds",
+        seqs=expand("results/diffsplice/{{model}}.sequences/isoformSwitchAnalyzeR_isoform_{type}.fasta", type=["AA", "nt"])
+    log:
+        "logs/diffsplice/{model}.init.log"
+    conda:
+        "../envs/isoform-switch-analyzer.yaml"
+    params:
+        model=get_model,
+        samples=["{unit.sample}-{unit.unit}".format(unit=unit) for unit in units.itertuples()],
```
dlaehnemann, line 16: I think this would conflict with the merging of samples into one file per sample, depending on how we resolve PR #16 .

.test/config/config.yaml
johanneskoester: That sounds good!

workflow/rules/common.smk
```diff
@@ -28,7 +28,8 @@ report: "../report/workflow.rst"
 
 wildcard_constraints:
     sample="|".join(samples.index),
-    unit="|".join(units["unit"])
+    unit="|".join(units["unit"]),
+    model="[^\.]+"
```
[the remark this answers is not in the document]
johanneskoester: Because we use dots as separators in filenames. The schema should be updated to contain the same regex for the model names.

workflow/scripts/sleuth-init.R
```diff
@@ -35,21 +35,36 @@ if(!is.null(snakemake@params[["exclude"]])) {
...
-                    c( sample, path, all_of(variables) )
-                )
+	        drop_na(c(sample, path, all_of(variables)))
+
+    primary_variable <- model[["primary_variable"]]
+    base_level <- model[["base_level"]]
+    # TODO migrate this to tidyverse
+    # Ensure that primary variable factors are sorted such that base_level comes first.
+    # This is important for fold changes, effect sizes to have the expected sign.
+    samples[, primary_variable] <- relevel(samples[, primary_variable, drop=TRUE], base_level)
```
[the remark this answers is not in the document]
johanneskoester: Excellent!

workflow/scripts/isoform-switch-analysis-init.R
johanneskoester: I think so.

PR #40, opened 2021-04-30
Miscellaneous fixes

tedil:
This PR aims to fix some issues I encountered when using the pipeline:
  - The logic in `rules/common.smk::is_single_end` was somewhat backwards
  - fgsea:
    - fgsea has an option to collapse enriched pathways, such that they become independent; this now also gets plotted
    - fgsea now only plots (at most) 1000 significant (w.r.t. fdr) pathways
    - pathways may have weird names, so resulting filenames might not be valid, i.e. saving to file failed in some cases; to fix that, replace unwanted characters with underscore
  - the diffexp-heatmap plot now uses gene names instead of transcript names (as suggested in the TODO)
  - do not load sleuth object twice in `plot-pca.R`
  - spia: if there are no significant genes, make an empty plot instead of failing
  - common.smk: 
    - allow `#` comments in samples.tsv + units.tsv
    - remove singular columns from samples dataframe (since every column is used in the analysis scripts at some point, and singular ones just make everything fail)

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -15,13 +15,20 @@ configfile: "config/config.yaml"
 
 validate(config, schema="../schemas/config.schema.yaml")
 
-samples = pd.read_csv(config["samples"], sep="\t", dtype=str).set_index(
+samples = pd.read_csv(config["samples"], sep="\t", dtype=str, comment="#").set_index(
     "sample", drop=False
 )
 samples.index.names = ["sample_id"]
+
+
+def drop_unique_cols(df):
+    return df.drop(df.nunique().loc[(df.nunique().values <= 1)].index, axis=1)
+
+samples = drop_unique_cols(samples)
+print(samples)
 validate(samples, schema="../schemas/samples.schema.yaml")
 
-units = pd.read_csv(config["units"], dtype=str, sep="\t").set_index(
+units = pd.read_csv(config["units"], dtype=str, sep="\t", comment="#").set_index(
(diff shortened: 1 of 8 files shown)


johanneskoester (approved):
LGTM!

workflow/rules/common.smk
```diff
@@ -57,16 +65,8 @@ def get_model(wildcards):
...
     """Determine whether unit is single-end."""
-    fq2_present = pd.isnull(units.loc[(sample, unit), "fq2"])
-    if isinstance(fq2_present, pd.core.series.Series):
-        # if this is the case, get_fastqs cannot work properly
-        raise ValueError(
-            f"Multiple fq2 entries found for sample-unit combination {sample}-{unit}.\n"
-            "This is most likely due to a faulty units.tsv file, e.g. "
-            "a unit name is used twice for the same sample.\n"
-            "Try checking your units.tsv for duplicates."
-        )
```
dlaehnemann, line 68: @tedil was it intentional to remove this check? I remember that we introduced this because multiple occurrences of the same `{sample}-{unit}` combination otherwise throws an incomprehensible error. So even if this might not be the best place to put it, we should check for this somewhere...

tedil: True, the check should be done somewhere, probably directly after reading samples.tsv and units.tsv.
The way this check was written didn't really work anyways, since fq2_present always is a pd.Series, irrespective of the cardinality of sample-unit combinations

PR #43, opened 2021-05-14
automatically download reference

dlaehnemann:
This also required adjusting a bunch of functions in `common.smk`, but our tests should catch any issues with that.

P.S.: I think I need a clean new checkout of this repo locally. Dunno where all these master branch merges and old commits come from...

### DIFF
--- a/workflow/rules/ref.smk
+++ b/workflow/rules/ref.smk
@@ -11,4 +11,3 @@ rule get_genome:
     cache: True
     wrapper:
         "0.74.0/bio/reference/ensembl-sequence"
-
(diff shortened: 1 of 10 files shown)


workflow/rules/ref.smk
```diff
@@ -0,0 +1,13 @@
+rule get_genome:
```
johanneskoester, line 1: Isn't this supposed to be called get_transcriptome?

johanneskoester (approved):

johanneskoester (changes requested):
see above

workflow/rules/ref.smk
tedil: ... and the output `"resources/transcriptome.fasta"`?

dlaehnemann: Oups. Copy-pasta...

johanneskoester (approved):

PR #45, opened 2021-07-28
Interactive volcano plots

tedil:
This PR adds *interactive* volcano plots. This allows displaying hover information such as gene symbols, qvalue, pvalue, signed pi value, transcript. It also uses q-values on the y-axis instead of p-values. Circle size scales with standard error of the beta value of the covariate. I tried using error bars, but that looks really weird if you have a lot of points.


workflow/Snakefile
```diff
@@ -20,6 +20,7 @@ include: "rules/ref.smk"
 include: "rules/quant.smk"
 include: "rules/diffexp.smk"
 include: "rules/enrichment.smk"
+include: "rules/plot.smk"
```
dlaehnemann, line 23: Do we really need another smk-file for this? Shouldn't this fit into `diffexp.smk`? Also, if we have a `plot.smk`, I would expect all rules for plots to be found there, so this could even be confusing.

workflow/rules/plot.smk
```diff
@@ -0,0 +1,20 @@
+rule interactive_plots:
+    input:
+        tsv="results/tables/diffexp/{model}.transcripts.diffexp.tsv",
+    output:
+        volcano_plot=directory(
+            report(
+                "results/plots/interactive/volcano/{model}/",
+                caption="../report/plot-volcano.rst",
+                category="QC",
```
dlaehnemann, line 9: As you pointed out in the PR message, this is probably not the best category for it. As it is an actual result and can generate multiple plots across multiple models, it could even go into its own `Volcano plot` category.

workflow/rules/common.smk
```diff
@@ -173,6 +173,7 @@ def all_input(wildcards):
             [
                 "results/plots/mean-var/{model}.mean-variance-plot.pdf",
                 "results/plots/volcano/{model}.volcano-plots.pdf",
```
dlaehnemann, line 175: I think we should remove the old volcano plot to avoid confusion. One thing to double-check, is that plotly enables an export of a figure into something like a PDF or JPEG.

dlaehnemann (commented):
Looks good, thanks a lot. A couple of smaller suggestions below.

workflow/rules/plot.smk
tedil: Or a general "interactive plots" category with a volcano subcategory. There are a lot of options, we just have to pick one ;)

workflow/Snakefile
tedil: Well, it's not expensive to have one more smk file ;)
I thought of moving all plot related rules to plot.smk, to have them all in one place. However, some plots are part/byproduct of e.g. sleuth routines, so perhaps that's not feasible.
Also, we might want to add more (interactive/custom) plots in the future. To go even more overboard with this, we might have a whole "plot" subdirectory of smk files for different parts of the workflow.

Anyways, for now I'll either move all plotting rules into plot.smk or move that single rule to diffexp.smk, probably the latter. Thoughts?

workflow/rules/common.smk
tedil: plotly + kaleido can generate those (and you can always hit the save button in the top right corner of html plots to get a png render of that plot; you just don't have much control about the png's resulting size in that case).
Do we replace those pdfs with plotly versions or do we remove them altogether?

tedil:
The volcano plots from sleuth really seem to be for QC only, since those use beta values vs qval *per covariate*, but not for the whole model. For the interactive plots, we want the whole model. We can stick with the last changes I made and use those instead of the sleuth ones for QC, but additionally provide the whole-model volcano plot in a different (sub)category in the snakemake report.
Also @fxwiegand has been so kind as to translate the plots to vega-lite, so we might as well use those (the only qualm I have with those is that vega's hover tooltips are somewhat hard to trigger).

dlaehnemann:
Yes, exactly. Thanks for keeping at this, even though I am so slow to respond during my parental leave.

I like the idea of replacing the sleuth plots, especially if it is with vega-lite plots (thanks @fxwiegand). This will make them so much more configurable in case we or collaborators want to twiddle some more.

And for the results, we definitely want the whole model qval volcano plots.

workflow/Snakefile
johanneskoester: For now, diffexp.smk seems to be the right place indeed.

johanneskoester (approved):

workflow/Snakefile
johanneskoester: Can you move it there? Then, this is ready for merging I think.

PR #55, opened 2022-02-01
adjust fgsea.R script to changes in fgsea() function interface

dlaehnemann:
Originally, the `fgsea()` function had an argument `nperm=` that determined the precision of p-value estimates. See for example this legacy documentation:
https://bioconductor.org/packages/3.10/bioc/vignettes/fgsea/inst/doc/fgsea-tutorial.html

Starting with bioconductor `3.11`, this precision is now directly specified via the `eps=` argument. See for example this version of the documentation:
https://bioconductor.org/packages/3.11/bioc/vignettes/fgsea/inst/doc/fgsea-tutorial.html

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -83,7 +83,7 @@ enrichment:
     # if activated, you need to provide a GMT file with gene sets of interest
     gene_sets_file: "resources/gene_sets/dummy.gmt"
     fdr_gene_set: 0.05
-    nperm: 100000
+    eps: 0.0
   spia:
     # tool is only run if set to `true`
     activate: false
--- a/.test/config/config.yaml
+++ b/.test/config/config.yaml
@@ -75,7 +75,7 @@ enrichment:
     activate: true
     # if activated, you need to provide a GMT file with gene sets of interest
     fdr_gene_set: 0.05
-    nperm: 10000
+    eps: 0.0001
   spia:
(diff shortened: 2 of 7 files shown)


config/config.yaml
```diff
@@ -83,7 +83,7 @@ enrichment:
     # if activated, you need to provide a GMT file with gene sets of interest
     gene_sets_file: "resources/gene_sets/dummy.gmt"
     fdr_gene_set: 0.05
-    nperm: 100000
+    eps: 0.0
```
tedil, line 86: Since this changes here, the config schema has to be changed as well. It appears there's not test that covers this case of an incorrect schema?

tedil (changes requested):

config/config.yaml
dlaehnemann: I just forgot to update the two relevant places for this in sync: `workflow/schemas/config.schema.yaml` AND `.test/config/config.yaml`.
If you forget both, the test won't complain... Oups.

dlaehnemann: But this should be resolved with the following 3 commits. Which shows that I also forgot to update a third place, that the linting now found...

dlaehnemann: Make it 4.

tedil (approved):

PR #62, opened 2022-10-07
feat: Support for 3-prime RNA sequencing 

manuelphilip:


config/config.yaml
```diff
@@ -26,6 +30,9 @@ resources:
   ontology:
     # gene ontology to download, used e.g. in goatools
     gene_ontology: "http://current.geneontology.org/ontology/go-basic.obo"
+  gencode:
+    gencodeBED: "https://sourceforge.net/projects/rseqc/files/BED/Human_Homo_sapiens/hg38_GENCODE.v38.bed.gz/download"
```
johanneskoester, line 34: What is this about?

config/config.yaml
```diff
@@ -43,23 +50,25 @@ diffexp:
...
+      #full: ~condition + batch_effect
+      reduced: '~1'
+      #reduced: '~1'# in case if it takes as numerical
       # Binary valued covariate that shall be used for fold change/effect size
       # based downstream analyses.
       primary_variable: condition
       # base level of the primary variable (will be considered as denominator 
       # in the fold change/effect size estimation).
-      base_level: untreated
+      base_level: Non_resp
```
johanneskoester, line 63: Let's not change the configuration example.

config/config.yaml
```diff
@@ -43,23 +50,25 @@ diffexp:
...
   # significance level to use for volcano, ma- and qq-plots
   sig-level:
     volcano-plot: 0.05
     ma-plot: 0.05
     qq-plot: 0.05
 
 diffsplice:
-  activate: true
+  activate: false
```
johanneskoester, line 71: Let's not change the config example.

config/config.yaml
```diff
@@ -118,8 +129,9 @@ params:
   # also, separate capitalised letter flags are required for adapters in
   # the reverse reads of paired end sequencing:
   # * https://cutadapt.readthedocs.io/en/stable/guide.html#trimming-paired-end-reads
-  cutadapt-se: ""
+  cutadapt-se: " "
```
johanneskoester, line 132:
```suggestion
  cutadapt-se: ""
```

config/config.yaml
```diff
@@ -118,8 +129,9 @@ params:
...
   # reasoning behind parameters:
+  #   For reads that are produced by 3’-end sequencing first 12 bases are removed based on accordance with Lexogen’s recommendation. More info: https://www.nature.com/articles/s41598-019-55434-x#Sec10
```
johanneskoester, line 134:
```suggestion
  #   For reads that are produced by 3’-end sequencing, depending on the protocol, it might be recommended to remove some leading bases (e.g. see https://www.nature.com/articles/s41598-019-55434-x#Sec10)
```

johanneskoester: I think this comment (with my changes) should be moved into the cutadapt-se part above.

config/samples.tsv
```diff
@@ -1,5 +1,5 @@
 sample	condition	batch_effect
-A	treated	batch1
-B	untreated	batch1
-C	treated	batch2
-D	untreated	batch2
+1738_17	SuperResp	batch1
+10435_17	Non_resp	batch1
+10_9448_IIIa	SuperResp	batch1
+15281_17_A	SuperResp	batch1
```
johanneskoester, line 5: Please don't change the example.

config/units.tsv
```diff
@@ -1,6 +1,8 @@
...
+1738_17	1	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/1738_17_S51_L001_R1_001.fastq.gz	
+10435_17	1	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/10435_17_S72_L001_R1_001.fastq.gz	
+10435_17	2	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/10435_17_S72_L002_R1_001.fastq.gz	
+10_9448_IIIa	1	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/10_9448_IIIa_S43_L001_R1_001.fastq.gz	
+10_9448_IIIa	2	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/10_9448_IIIa_S43_L002_R1_001.fastq.gz	
+15281_17_A	1	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/15281_17_A_S44_L001_R1_001.fastq.gz	
+15281_17_A	2	75	1	/home/derma/RNAseq_Essen/20220411_RNAseq_SR_Essen/220405_VH00142_35_AAALH2NHV_BC22_06_Sucker/Analysis/1/Data/fastq/15281_17_A_S44_L002_R1_001.fastq.gz
```
johanneskoester, line 8: Please don't change the example.

workflow/Snakefile
```diff
@@ -21,8 +21,12 @@ include: "rules/quant.smk"
...
 include: "rules/enrichment.smk"
+include: "rules/datavzrd.smk"
 
 
 rule all:
     input:
         all_input,
+        "results/QC/QC_plot.html" 
+        if config["experiment"]["3-prime-rna-seq"]["activate"]
+        else all_input,
```
johanneskoester, line 32: Rather than handling here, I'd add this to all_input at the place where that is generated.

workflow/envs/heatmap.yaml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - conda-forge
+  - bioconda
```
johanneskoester, line 3:
```suggestion
  - bioconda
  - nodefaults
```

workflow/envs/pysam.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - defaults
+  - bioconda
```
johanneskoester, line 4:
```suggestion
  - bioconda
  - nodefaults
```

workflow/envs/r-fasta.yaml
```diff
@@ -0,0 +1,6 @@
+channels:
+  - conda-forge
+  - bioconda
```
johanneskoester, line 3:
```suggestion
  - bioconda
  - nodefaults
```

workflow/envs/spia.yaml
```diff
@@ -7,3 +7,4 @@ dependencies:
   - bioconductor-spia =2.44
   - bioconductor-graphite =1.38
   - r-tidyverse =1.3
+  - r-ggplot2
```
johanneskoester, line 10: needs a version

workflow/report/plot-QC.rst
```diff
@@ -0,0 +1,2 @@
+**Quality control plot**,Quality control plot of all samples as step chart. The left plot shows the difference between transcripts length and read start postion upto 20,000 base pairs (x-axis).
+The right plot show the full length transcript (x-axis).
```
johanneskoester, line 2:
```suggestion
Histogram of read start positions of all samples as step chart. The left plot shows the first 20,000 bases from 3' end (x-axis).
The right plot shows the full length transcript (x-axis).
```

workflow/report/plot-heatmap.rst
```diff
@@ -1,3 +1,3 @@
 
-Heatmap of top 20 differentially expressed transcripts using the model ``{{ snakemake.params.model["full"] }}``, computed with sleuth.
+Heatmap of top 50 differentially expressed transcripts_genes using the model ``{{ snakemake.params.model["full"] }}``, from sleuth.
```
johanneskoester, line 2: I think apart from the number, no change is required in above text.

workflow/report/plot-pathway.rst
```diff
@@ -0,0 +1,4 @@
+**Pathway enrichment** performed with SPIA, using the model ``{{ snakemake.config["diffexp"]["models"][snakemake.wildcards.model]["full"] }}``.
+
+The plot is filtered based on ``pGFdr`` <0.05   On the x-axis contains the ``tA`` is the observed total perturbation accumulation in the pathway and y-axis contains the pathways. Bars are coloured based on the ``Status`` gives the direction in which the pathway is perturbed (``activated`` or ``inhibited``) (also see the `SPIA docs <https://rdrr.io/bioc/SPIA/man/spia.html>`_).
+
```
johanneskoester, line 4: The plot is gone, isn't it? We just have the datavzrd report for spia? Then, this explanation can be adapted for the datavzrd case and renamed accordingly.

workflow/rules/common.smk
```diff
@@ -120,7 +129,7 @@ def get_bioc_pkg_path(wildcards):
 
 def kallisto_params(wildcards, input):
     extra = config["params"]["kallisto"]
-    if len(input.fq) == 1:
+    if len(input.fq) == 1 or config["experiment"]["3-prime-rna-seq"]["activate"]:
```
johanneskoester, line 132: With 3-prime, samples should be single end, right? Hence, the "or" case is not required here.

workflow/rules/common.smk
```diff
@@ -129,12 +138,11 @@ def kallisto_params(wildcards, input):
         extra += " --fusion"
     return extra
 
-
 def all_input(wildcards):
     """
-    Function defining all requested inputs for the rule all (below).
+    #Function defining all requested inputs for the rule all (below).
```
johanneskoester, line 143:
```suggestion
    Function defining all requested inputs for the rule all (below).
```

workflow/rules/common.smk
```diff
@@ -280,4 +291,25 @@ def all_input(wildcards):
...
+        wanted_input.extend(
+            expand("results/kallisto_cds/{unit.sample}-{unit.unit}",unit=units.itertuples())
+        )
+        wanted_input.extend(
+            expand("results/canonical_reads/{unit.sample}-{unit.unit}.fastq",unit=units.itertuples())
+        )
+        if config["experiment"]["3-prime-rna-seq"]["plot-qc"] != "all":
+            wanted_input.extend(
+                expand("results/QC/{ind_transcripts}.QC_plot.html",ind_transcripts=config["experiment"]["3-prime-rna-seq"]["plot-qc"])
+            )
```
johanneskoester, line 314: Please check which of these really need to be global targets. *aligned.txt probably not, the kallisto output folders neither, nor the fastqs or the read_names.txt

workflow/rules/datavzrd.smk
```diff
@@ -0,0 +1,29 @@
...
+        # files required for rendering the given configs
+        spia_results="results/tables/pathways/{model}.pathways.tsv",
+        barplot="results/plots/pathways/{model}.pathways.html",
+    output:
+        report(
+            directory("results/datavzrd-reports/{type}.{model}"),
+            htmlindex="index.html",
+            # see https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html
+            # for additional options like caption, categories and labels
+        ),
```
johanneskoester, line 25: Please add caption, category, etc. You can follow what has been done for the spia tsv output.

workflow/rules/diffexp.smk
```diff
@@ -262,6 +242,42 @@ rule logcount_matrix:
         "../scripts/sleuth-to-matrix.R"
 
 
+rule get_heatmap:
+    input:
+        Sleuth_logcountmatrix_file="results/tables/logcount-matrix/{model}.logcount-matrix.tsv",
```
johanneskoester, line 247: please adhere to the python style guide (non-capital variable and argument names). Just `logcount_matrix=` is sufficient.

workflow/rules/diffexp.smk
```diff
@@ -262,6 +242,42 @@ rule logcount_matrix:
...
+        "logs/plots/diffexp-heatmap/{model}.diffexp-heatmap.log",
+    conda:
+        "../envs/heatmap.yaml"
+    script:
+        "../scripts/get_heatmap.R"
+
+
+rule get_heatmap_for_predefine_genes:
+    input:
+        Sleuth_logcountmatrix_file="results/tables/logcount-matrix/{model}.logcount-matrix.tsv",
```
johanneskoester, line 269: naming, see above

workflow/rules/diffexp.smk
```diff
@@ -262,6 +242,42 @@ rule logcount_matrix:
         "../scripts/sleuth-to-matrix.R"
 
 
+rule get_heatmap:
```
johanneskoester, line 245: Rename rule into plot_diffexp_heatmap and also rename the script accordingly.

workflow/rules/enrichment.smk
```diff
@@ -40,10 +40,29 @@ rule spia:
         "../envs/spia.yaml"
     log:
         "logs/tables/pathways/{model}.spia-pathways.log",
-    threads: 16
+    threads: 38
     script:
         "../scripts/spia.R"
 
+rule plot_spia_pathways:
```
johanneskoester, line 47: This rule and script can be removed as we have converted the contents into datavzrd.

workflow/rules/quant.smk
```diff
@@ -1,28 +1,141 @@
-rule kallisto_index:
+rule kallisto_cds_index:
```
johanneskoester, line 1: Since some time, there are Snakemake wrappers for kallisto quant and kallisto index: https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/kallisto.html

Please change all these rules here to use those wrappers instead of manually defined shell commands and conda envs.

workflow/rules/quant.smk
```diff
@@ -1,28 +1,141 @@
...
+        log:
+            "results/logs/kallisto_3prime/quant/{sample}-{unit}.log",
+        params:
+            extra=kallisto_params,
+        conda:
+            "../envs/kallisto.yaml"
+        shell:
+            "kallisto quant -i {input.idx} -o {output.kallisto_folder} {params.extra} {input.fq} 2> {log}"
+
+if config["experiment"]["3-prime-rna-seq"]["activate"]:
```
johanneskoester, line 60: Above, there is the same if already. You can remove this one.
```suggestion
```

workflow/rules/trim.smk
```diff
@@ -13,15 +13,69 @@ rule cutadapt_pe:
...
+            get_fastqs,
+        output:
+            fastq="results/trimmed/{sample}-{unit}.fastq.gz",
+            qc="results/trimmed/{sample}-{unit}.qc.txt",
+        params:
+            "{}".format(config["params"]["cutadapt-se"]),
+        log:
+            "results/logs/cutadapt/{sample}-{unit}.log",
+        wrapper:
+            "0.31.1/bio/cutadapt/se"
```
johanneskoester, line 70: Why 3 cutadapt steps?

workflow/scripts/get-3prime-seqs.R
```diff
@@ -0,0 +1,32 @@
...
+    "transcript_is_canonical"), mart = ensembl)
+read_length <- fromJSON(file = snakemake@input[["read_length"]])
+canonical_ids <- subset(get_canonical_transcripts, transcript_is_canonical == 1)
+seq <- getSequence(id = canonical_ids$ensembl_transcript_id_version,
+    type = "ensembl_transcript_id_version", seqType = "3utr",
+    mart = ensembl)
+seq_fil <-
+    seq %>% filter(`3utr` != "Sequence unavailable")
+seq_fil <- seq_fil %>% mutate(`3utr` = str_remove_all(`3utr`, "AAA"))
+seq_fil <- seq_fil %>% mutate(`3utr` = str_remove_all(`3utr`, "TTT"))
```
johanneskoester, line 25: Please add some comments and some space between logical blocks in order to increase readability.

workflow/scripts/get_3prime-seqs.py
```diff
@@ -0,0 +1,17 @@
...
+
+three_prime_output_file = open(snakemake.output[0], "w")
+
+from Bio import SeqIO
+for seq_record in SeqIO.parse(snakemake.input["ref_fasta"], "fasta"):
+    transcript_location = seq_record.description.split(" ")[0]
+    transcript_id = transcript_location.split("_")[0]
+    print(">",transcript_id, sep = "", file = three_prime_output_file)
+    print(seq_record.seq, file = three_prime_output_file)
+three_prime_output_file.close()
```
johanneskoester, line 17: Idiomatic Python is to use `with open(snakemake.output[0], "w") as three_prime_output_file:` instead of manually closing below.

workflow/scripts/get_aligned_reads.py
```diff
@@ -0,0 +1,33 @@
...
+#reads aligned to forward strand
+forward_strand = merge_data.loc[merge_data['Strand'] == '1']
+forward_strand[sample_name + '_forward_strand'] = forward_strand['LN'] - forward_strand['Start']
+aligned_reads = forward_strand.loc[forward_strand.groupby(['read_name','read'])[sample_name + '_forward_strand'].idxmin()]
+#reads aligned to reverse strand
+reverse_strand = merge_data.loc[merge_data['Strand'] == '-1']
+read_min = reverse_strand.loc[reverse_strand.groupby(['read_name','read'])['Start'].idxmin()]
+aligned_reads = pd.concat([aligned_reads, read_min])
+final_aligned_reads = aligned_reads.merge(canonical_ids, on="Transcript")
+final_aligned_reads['read_name'].to_csv(snakemake.output['aligned_read_names'], sep='\t', index=False)
```
johanneskoester, line 33: Please add some spacing and additional comments to increase readability.

workflow/scripts/get_pathway_barplot.R
```diff
@@ -0,0 +1,20 @@
+log <- file(snakemake@log[[1]], open = "wt")
```
johanneskoester, line 1: Script can be removed in favor of the datavzd report.

workflow/scripts/get_pathway_barplot.py
```diff
@@ -0,0 +1,15 @@
+import altair as alt
```
johanneskoester, line 1: Script can be removed in favor of the datavzd report.

workflow/scripts/sleuth-diffexp.R
```diff
@@ -191,7 +191,11 @@ write_results <- function(so, mode, output, output_all) {
 
     # add sample expressions
     all <- all %>% left_join(as_tibble(sleuth_to_matrix(so, "obs_norm", "est_counts"), rownames="target_id"))
-    write_tsv(all, path = output, quote_escape = "none")
+    #write_tsv(all, path = output, quote_escape = "none")
+    #Warning message:
+    #The `quote_escape` argument of `write_delim()` is deprecated as of readr 2.0.0.
+    #Please use the `escape` argument instead.
```
johanneskoester, line 197:
```suggestion
```

johanneskoester (changes requested):
Epic work, thanks a lot! I have various requests below.

workflow/Snakefile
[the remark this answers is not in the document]
manuelphilip: Done.

workflow/envs/aligned_pos.yaml
```diff
@@ -0,0 +1,5 @@
+channels:
+  - bioconda
+  - conda-forge
+dependencies:
+  - samtools
```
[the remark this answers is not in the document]
manuelphilip: Noted. Done.

config/config.yaml
```diff
@@ -1,5 +1,9 @@
 samples: config/samples.tsv
 units: config/units.tsv
+experiment:
+  3-prime-rna-seq:
+    activate: true
+    plot-qc: all
```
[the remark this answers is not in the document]
manuelphilip: Added comment as mentioned below
"# If set to `true`, this option allows the workflow to analyse 3-prime RNA seq data obtained from Quantseq protocol by Lexogen. For more information https://www.lexogen.com/quantseq-3mrna-sequencing/"

config/config.yaml
manuelphilip: This was added for testing purposes and is now deleted.

config/config.yaml
```diff
@@ -43,23 +50,25 @@ diffexp:
...
   # model for sleuth differential expression analysis
   models:
-    model_X:
-      full: ~condition + batch_effect
-      reduced: ~batch_effect
+    model_Condition:
+      full: ~condition
+      #full: ~condition + batch_effect
+      reduced: '~1'
+      #reduced: '~1'# in case if it takes as numerical
```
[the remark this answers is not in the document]
manuelphilip: Noted

config/config.yaml
manuelphilip: Noted

config/config.yaml
manuelphilip: Noted

config/config.yaml
```diff
@@ -108,7 +117,9 @@ plot_vars:
   sig_level: 0.1
 
 params:
-  kallisto: "-b 100"
+    #For reads that are produced by 3’-end sequencing, the --single-overhang option does not discard 
+    #reads where the expected fragment size goes beyond the transcript start
+  kallisto: "-b 100 --single-overhang --pseudobam"
```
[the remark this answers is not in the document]
manuelphilip: Done. Changed in the common.smk ` kallisto_params` definition

config/samples.tsv
manuelphilip: Noted

config/units.tsv
manuelphilip: Noted

workflow/Snakefile
manuelphilip: Noted. Done

workflow/envs/spia.yaml
manuelphilip: Noted. Done

workflow/report/plot-heatmap.rst
manuelphilip: Noted

workflow/rules/common.smk
manuelphilip: Noted, Changed

workflow/rules/common.smk
manuelphilip: Noted. Changed to required ones.

workflow/rules/datavzrd.smk
manuelphilip: Noted

workflow/rules/diffexp.smk
manuelphilip: Noted

workflow/rules/diffexp.smk
manuelphilip: Noted

workflow/rules/diffexp.smk
manuelphilip: Done

workflow/rules/diffexp.smk
```diff
@@ -262,6 +242,42 @@ rule logcount_matrix:
...
+            "primary_variable"],
+    log:
+        "logs/plots/diffexp-heatmap/{model}.diffexp-heatmap.log",
+    conda:
+        "../envs/heatmap.yaml"
+    script:
+        "../scripts/get_heatmap.R"
+
+
+rule get_heatmap_for_predefine_genes:
```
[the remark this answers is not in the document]
manuelphilip: Done

workflow/rules/quant.smk
manuelphilip: Noted. Changed

workflow/rules/quant.smk
manuelphilip: Noted.

workflow/rules/trim.smk
manuelphilip: In the Quantseq FAQ pages, they mentioned 3 cutadapt steps for quantseq 3 prime data
"https://faqs.lexogen.com/faq/what-is-the-adapter-sequence-i-need-to-use-for-t-1"

workflow/scripts/get-3prime-seqs.R
```diff
@@ -0,0 +1,32 @@
...
+#seq_fil <- seq_fil %>% mutate(cdna = str_remove_all(cdna, "AAA"))
```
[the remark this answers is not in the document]
manuelphilip: Noted. Removed

workflow/scripts/get-3prime-seqs.R
```diff
@@ -0,0 +1,32 @@
...
+seq_fil$`3utr` <- str_sub(seq_fil$`3utr`,
+    start = -read_length)
+#unava_seq_fil <-
+    #seq %>% filter(`3utr` != "Sequence unavailable")
+    # if unavailable sequences are to be removed.
```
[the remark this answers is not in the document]
manuelphilip: Yes, this can be removed.

workflow/scripts/get_3prime-seqs.py
manuelphilip: Thank you. That was a learning part.

workflow/scripts/get_aligned_reads.py
manuelphilip: This script has been removed now as it is no longer needed

workflow/scripts/get_canonical_ids.R
```diff
@@ -0,0 +1,24 @@
...
+    getBM(attributes = c("ensembl_transcript_id_version",
+        "transcript_is_canonical", "transcript_mane_select", "chromosome_name"),
+            mart = ensembl)
+#transcript_mane_select- manually curated primary transcript as it has been encoded in NCBI
+canonical_ids <- get_transcripts_ids %>%
+    select(ensembl_transcript_id_version, transcript_is_canonical,
+        transcript_mane_select,chromosome_name) %>%
+            filter(!str_detect(chromosome_name, "patch|PATCH")) %>%
+                filter(str_detect(transcript_mane_select, "")) %>%
+                    subset(transcript_is_canonical == 1)
```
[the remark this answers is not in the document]
manuelphilip: Yes, as mentioned previously, We keep this script instead of `get-3prime-seqs.R`.

workflow/scripts/get_histogram.py
```diff
@@ -0,0 +1,107 @@
+from turtle import title
```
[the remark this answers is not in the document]
manuelphilip: I changed the name to `plot-sample-QC-histogram.py`

workflow/scripts/get_pathway_barplot.R
manuelphilip: Noted.

workflow/scripts/get_pathway_barplot.py
manuelphilip: Noted

workflow/report/plot-pathway.rst
manuelphilip: Yes. Noted.

workflow/scripts/get_histogram.py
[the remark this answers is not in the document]
johanneskoester: Can you change it to `plot-3prime-qc-histogram.py`?

workflow/scripts/plot_ind-transcripts_histogram.py
```diff
@@ -0,0 +1,113 @@
+from turtle import title
```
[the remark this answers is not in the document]
johanneskoester: But still, I am quite sure that this is just a matter of providing a special parameter in the script isn't it? Most of the code looks identical.

config/config.yaml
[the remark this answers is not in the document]
johanneskoester: isn't this still a change to before?

config/config.yaml
johanneskoester: Still different from before.

config/config.yaml
johanneskoester: Still different from before

config/samples.tsv
```diff
@@ -1,5 +1,5 @@
 sample	condition	batch_effect
-A	treated	batch1
-B	untreated	batch1
-C	treated	batch2
-D	untreated	batch2
+A   treated batch1
+B   untreated   batch1
+C   treated batch2
+D   untreated   batch2
```
johanneskoester, line 5: Please restore the old sheet (after your changes, it contains spaces instead of tabs, but this is a .tsv file).

workflow/rules/trim.smk
```diff
@@ -13,15 +13,75 @@ rule cutadapt_pe:
...
+            fastq="results/trimmed/{sample}-{unit}.2.1.fastq.gz",
+        output:
+            fastq="results/trimmed/{sample}-{unit}.fastq.gz",
+            qc="results/trimmed/{sample}-{unit}.qc.txt",
+        params:
+            adapters=lambda w: str("-m 20 -O 20 -g \"""r1adapter=AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=20\""" --discard-trimmed"),
+        log:
+            "results/logs/cutadapt/{sample}-{unit}.log",
+        wrapper:
+            "v1.14.1/bio/cutadapt/se"
```
johanneskoester, line 58: For these 3 steps, I need a detailed explanation, e.g.
1. Why do we trim twice with the same adapter, but different overlaps?
2. The first iteration trims only poly-A and something with the quality? Can't this be done along with any of the other iterations?

In any case, a comment with a rationale and human readable description has to be provided for each iteration.

workflow/rules/trim.smk
```diff
@@ -13,15 +13,75 @@ rule cutadapt_pe:
...
+            qc="results/trimmed/{sample}-{unit}.qc.txt",
+        params:
+            "{}".format(config["params"]["cutadapt-se"]),
+        log:
+            "results/logs/cutadapt/{sample}-{unit}.log",
+        wrapper:
+            "0.31.1/bio/cutadapt/se"
+
+
+rule get_all_fastqs:
```
johanneskoester, line 78: This rule gets the max read length, right? It should be named like that.

workflow/scripts/goatools-go-enrichment-analysis.py
```diff
@@ -52,6 +51,26 @@
...
+    
+
+temp = goeaobj.ns2objgoea['CC'].assoc
+for item in temp:
+    GO_items += temp[item]
+    
+
+temp = goeaobj.ns2objgoea['MF'].assoc
+for item in temp:
+    GO_items += temp[item]
```
johanneskoester, line 69: Simplification, please check:
```suggestion
go_items = [val for cat in ["BP", "CC", "MF"] for item, val in goeaobj.nsobjgoea[cat].assoc.items()]
```
Also, note that variable names in Python have to be in snake_case (all lower).

johanneskoester (changes requested):

workflow/scripts/plot_ind-transcripts_histogram.py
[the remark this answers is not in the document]
manuelphilip: Noted. The script is changed and now instead of two scripts, we have a single script for both `all` transcripts and `single` transcripts.

workflow/scripts/get_histogram.py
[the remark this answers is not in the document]
manuelphilip: Noted. Changed, the file name.

workflow/report/plot-QC.rst
```diff
@@ -0,0 +1,2 @@
+Histogram of read start positions of all samples as step chart. The left plot shows the first 20,000 bases from 3' end (x-axis).
+The right plot shows the full length transcript (x-axis).
```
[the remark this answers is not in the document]
manuelphilip: Noted. Changed file name

workflow/rules/common.smk
```diff
@@ -280,4 +291,12 @@ def all_input(wildcards):
             )
         )
 
-    return wanted_input
+    if config["experiment"]["3-prime-rna-seq"]["activate"]:
+        wanted_input.extend(
+            expand("results/plots/QC/{model}.QC-plot.html",model=config["diffexp"]["models"])
```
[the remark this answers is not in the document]
manuelphilip: Noted. Changed.

workflow/rules/quant.smk
```diff
@@ -1,28 +1,259 @@
...
+            else "resources/transcriptome.cdna.fasta",                
     output:
-        "results/kallisto/transcripts.idx",
+        index="results/kallisto_cds/transcripts.idx",
     log:
-        "results/logs/kallisto/index.log",
+        "results/logs/kallisto_cds/index.log",
+    threads: 1
+    wrapper:
+        "v1.17.4/bio/kallisto/index"
```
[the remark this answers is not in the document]
manuelphilip: Noted. Updated the code

workflow/rules/quant.smk
```diff
@@ -1,28 +1,259 @@
...
+        index="results/kallisto_cds/transcripts.idx",
+    output:
+        kallisto_folder=directory("results/kallisto_cds/{sample}-{unit}"),
+    log:
+        "results/logs/kallisto_cds/quant/{sample}-{unit}.log",
+    params:
+        extra=kallisto_params,
+    threads: 5
+    wrapper:
+        "v1.17.4/bio/kallisto/quant"
```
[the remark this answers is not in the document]
manuelphilip: Noted. Updated the code

workflow/rules/quant.smk
```diff
@@ -1,28 +1,259 @@
...
+    output:
+        canonical_mapped_bam=temp("results/canonical_mapped_bam/{sample}-{unit}.canonical.mapped.bam"),
+    log:
+        "results/logs/canonical_mapped_bam/{sample}-{unit}.canonical-mapped-bam.log",
     conda:
-        "../envs/kallisto.yaml"
+        "../envs/samtools.yaml"
     shell:
-        "kallisto index -i {output} {input} 2> {log}"
+        "samtools view -h {input.mapped_bam} |  cut -f1-12 | grep -f {input.canonical_ids} > {output.canonical_mapped_bam}  2> {log}"
```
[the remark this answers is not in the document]
manuelphilip: I am taking columns up to the first 12 as the last column contains alternative aligned IDs which the piped `grep` will consider.

workflow/rules/quant.smk
```diff
@@ -1,28 +1,259 @@
...
+        canonical_mapped_bam="results/canonical_mapped_bam/{sample}-{unit}.canonical.mapped.bam",
+        canonical_mapped_3prime_pos="results/mapped_3prime_bam/{sample}-{unit}.canonical.mapped.3prime_pos.txt",
+    output:
+        canonical_mapped_3prime_bam=temp("results/canonical_3prime_mapped_bam/{sample}-{unit}.canonical.3prime_mapped.bam"),
+    log:
+        "results/logs/canonical_3prime_mapped_bam/{sample}-{unit}.canonical.3prime_mapped.log",
+    conda:
+        "../envs/samtools.yaml"
+    shell:
+        "samtools view -R {input.canonical_mapped_3prime_pos} {input.canonical_mapped_bam} -o {output.canonical_mapped_3prime_bam}  2> {log}"
```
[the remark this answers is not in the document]
manuelphilip: Yes. http://www.htslib.org/doc/samtools-view.html

workflow/rules/trim.smk
manuelphilip: Noted. Changed to `max_read_length`

workflow/rules/trim.smk
manuelphilip: I added the adapter options based on the Lexogen documentation. 
https://faqs.lexogen.com/faq/what-is-the-adapter-sequence-i-need-to-use-for-t-1
As I understood, they have used 3 Cutadapt options without specifying what each iteration does. However, I will add the information based on the options used by referring to the Cutadapt manual.

workflow/scripts/goatools-go-enrichment-analysis.py
manuelphilip: Noted. Changed

workflow/rules/datavzrd.smk
```diff
@@ -0,0 +1,60 @@
...
+    input:
+        config="results/datavzrd/{type}/{model}.yaml",
+        transcripts="results/tables/diffexp/{model}.transcripts.diffexp.tsv",
+        genes_aggregated="results/tables/diffexp/{model}.genes-aggregated.diffexp.tsv",
+        genes_representative="results/tables/diffexp/{model}.genes-representative.diffexp.tsv",
+        
+    output:
+        report(
+            directory("results/diffexp-reports/{type}.{model}"),
+            htmlindex="index.html",
```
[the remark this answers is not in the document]
manuelphilip: Noted. Added caption and category

workflow/rules/enrichment.smk
manuelphilip: Noted. removed

config/config.yaml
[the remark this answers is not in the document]
manuelphilip: Changed.

config/config.yaml
manuelphilip: Noted. Changed

config/config.yaml
manuelphilip: Changed

config/samples.tsv
manuelphilip: Done.

workflow/scripts/plot_heatmap_predefined.R
```diff
@@ -0,0 +1,27 @@
...
+
+predefine_genelist <- read.table(snakemake@input[["predef_genelist"]],
+   sep = " ", header = TRUE)
+
+selectedgenes <-
+    sleuth_file %>% filter(rownames(sleuth_file)
+     %in% predefine_genelist$Selected_genes)
+
+print(selectedgenes)
+png(snakemake@output[["predefgene_png"]], w = 1500, h = 1500, pointsize = 200)
```
johanneskoester, line 25: 1. Never write to PNG. USe PDF instead, since that is scalable, whereas PNG becomes blurry once you zoom in (as it is a raster graphic).
2. This script is mostly the same as plot_diffexp_heatmap. Both should be joined and parameterized instead.

workflow/scripts/plot-3prime-qc-histogram.py
```diff
@@ -0,0 +1,158 @@
+from turtle import title
```
johanneskoester, line 1: I thought you had merged this and the other script for the QC histogram. Did you just forget to delete one of them?

workflow/rules/ref.smk
```diff
@@ -13,7 +12,59 @@ rule get_transcriptome:
...
+    rule get_canonical_transcripts:
+        input:
+            fasta="resources/transcriptome.3prime.fasta",
+            canonical_ids="resources/canonical_ids.csv",
+        output:
+            "resources/transcriptome_clean.3prime.fasta",
+        conda:
+            "../envs/get_canonical_ids.yaml"
+        shell:
+            """bioawk -cfastx 'BEGIN{{while((getline k <"{input.canonical_ids}")>0)i[k]=1}}{{if(i[$name])print ">"$name"\\n"$seq}}' {input.fasta} > {output}"""
```
johanneskoester, line 67:
```suggestion
            "bioawk -cfastx "
            "'BEGIN{{while((getline k <"{input.canonical_ids}")>0)i[k]=1}}{{if(i[$name])print ">"$name"\\n"$seq}}' "
            "{input.fasta} > {output}"
```

johanneskoester: Further, please add a log file.

workflow/rules/quant.smk
```diff
@@ -1,28 +1,336 @@
...
+    log:
+        "results/logs/QC/{sample}-{unit}.sorted.index.log",
+    params:
+        extra="",  # optional params string
+    threads: 4  # This value - 1 will be sent to -@
+    wrapper:
+        "v1.18.3/bio/samtools/index"
+
+
+if is_3prime_experiment and config["experiment"]["3-prime-rna-seq"]["plot-qc"] != "all":
```
johanneskoester, line 216: Please move these into a separate file called `qc_3prime.smk`. Further, the rules are very similar. You can keep the transcript specific version you have below, and for the "all" version, you can simplify the definition via a use rule statement:

```python
use rule get_selected_transcripts_aligned_read_bins as get_aligned_read_bins with:
    output:
        # and define diverging output here
```

workflow/rules/quant.smk
```diff
@@ -1,28 +1,336 @@
...
+        "results/logs/kallisto_{type}/quant/{sample}-{unit}.log",
     params:
         extra=kallisto_params,
+    threads: 5
+    wildcard_constraints:
+        type="cdna|3prime",
+    wrapper:
+        "v1.17.4/bio/kallisto/quant"
+
+
```
johanneskoester, line 33: Move the 3prime specific quant rules below into a separate file quant_3prime.smk

workflow/rules/ref.smk
```diff
@@ -13,7 +12,59 @@ rule get_transcriptome:
         type="cdna|cds|ncrna",
     cache: "omit-software"
     wrapper:
-        "0.74.0/bio/reference/ensembl-sequence"
+        "v1.7.1/bio/reference/ensembl-sequence"
+
+
+if is_3prime_experiment:
```
johanneskoester, line 18: Move this into a separate file ref_3prime.smk

workflow/envs/r-fasta.yaml
```diff
@@ -0,0 +1,6 @@
...
+  - nodefaults
+dependencies:
+  - biopython =1.79
```
johanneskoester, line 6: PLease rename this into biopython.yaml

workflow/resources/datavzrd/diffexp-template.yaml
```diff
@@ -0,0 +1,280 @@
...
+                }
+        target_id:
+          display-mode: normal
+          link-to-url: "http://www.ensembl.org/Homo_sapiens/Gene/Summary?g={target_id}"
+        pval:
+          plot:
+            bars:
+              scale: linear
+              domain:
+                - 0.0000000000000000000000000000000000000000000000000000000000000000001
```
johanneskoester, line 80: Does scientific notation not work here? Also, why not just from 0.0 to 1.0?

workflow/rules/trim.smk
```diff
@@ -6,22 +6,112 @@ rule cutadapt_pe:
...
+    #   * Quality score<20 is trimmed, based on NextSeq/NovaSeq 2-color SBS system
+    #   * `-a A{18}AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=3;max_error_rate=0.100000`
+    #   * Trims the polyA-stretch + adapter at the 3' end with minimum overlap 3bp, maximum error of 1bp every 10bp
+
+    # Rule cutadapt3 removes final set of adapters from 5'end and discards reads based on trimmed read length
+    # reasoning behind parameters:
+    #   * `-m 20`:
+    #   * Discards any read under 20bp by -m 20
+    #   * `-O 20 -g r1adapter=AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=20 --discard-trimmed"`
+    #   * Trim the 5' end of adapter with miminum overlap of 20bp and discard its read by --discard-trimmed
```
johanneskoester, line 40: THis comes from lexogene, right?

1. Very important: please cite the page/manual you have this from here.
2. Move these rules into trim_3prime.smk

workflow/scripts/spia.R
```diff
@@ -46,5 +46,33 @@ if(nrow(sig_genes) == 0) {
...
+    names(res_reorder)[names(res_reorder) == "pGFWER"] <- "Combined Bonferroni p-values"
+    names(res_reorder)[names(res_reorder) == "pGFdr"] <- "Combined FDR"
+    names(res_reorder)[names(res_reorder) == "tA"] <- "total perturbation accumulation"
+    names(res_reorder)[names(res_reorder) == "pSize"] <- "number of genes on the pathway"
+    names(res_reorder)[names(res_reorder) == "NDE"] <- "number of DE genes per pathway"
+    names(res_reorder)[names(res_reorder) == "pG"] <- "Combined p-value"
+    names(res_reorder)[names(res_reorder) == "pPERT"] <- "p-value to observe a total accumulation"
+    names(res_reorder)[names(res_reorder) == "pNDE"] <- "p-value for at least NDE genes"
+    write_tsv(res_reorder, snakemake@output[["table"]])
+    }else {
```
johanneskoester, line 69: please ensure that this is properly formatted (e.g. using styleR or checking the R style guide manually).

workflow/scripts/spia.R
```diff
@@ -46,5 +46,33 @@ if(nrow(sig_genes) == 0) {
...
+    res_reorder <- dplyr::select(final_res, Name, Status,
+        pGFdr, tA, pSize, NDE, pNDE, pGFWER, pPERT, pG, Ids)
+    names(res_reorder)[names(res_reorder) == "pGFWER"] <- "Combined Bonferroni p-values"
+    names(res_reorder)[names(res_reorder) == "pGFdr"] <- "Combined FDR"
+    names(res_reorder)[names(res_reorder) == "tA"] <- "total perturbation accumulation"
+    names(res_reorder)[names(res_reorder) == "pSize"] <- "number of genes on the pathway"
+    names(res_reorder)[names(res_reorder) == "NDE"] <- "number of DE genes per pathway"
+    names(res_reorder)[names(res_reorder) == "pG"] <- "Combined p-value"
+    names(res_reorder)[names(res_reorder) == "pPERT"] <- "p-value to observe a total accumulation"
+    names(res_reorder)[names(res_reorder) == "pNDE"] <- "p-value for at least NDE genes"
```
johanneskoester, line 67: Please use dplyr for the renaming: https://dplyr.tidyverse.org/reference/rename.html

johanneskoester (changes requested):

workflow/scripts/plot_heatmap_predefined.R
manuelphilip: Noted. Changed.

workflow/scripts/plot-3prime-qc-histogram.py
manuelphilip: Yes. Now we have only one script. I deleted the older script.

workflow/rules/quant.smk
manuelphilip: Noted. Changed.

workflow/rules/quant.smk
manuelphilip: Noted. Changed

workflow/rules/ref.smk
manuelphilip: Moved.

workflow/envs/r-fasta.yaml
manuelphilip: Noted.

workflow/scripts/spia.R
manuelphilip: Noted. Updated

workflow/scripts/spia.R
manuelphilip: Noted.

workflow/resources/datavzrd/diffexp-template.yaml
manuelphilip: This is now updated. I guess in the previous version, there was an issue with the bar scale.

workflow/rules/ref.smk
manuelphilip: Noted. Updated the command into multiple lines

workflow/rules/trim.smk
manuelphilip: Noted.

config/config.yaml
```diff
@@ -124,7 +125,10 @@ params:
   # also, separate capitalised letter flags are required for adapters in
   # the reverse reads of paired end sequencing:
   # * https://cutadapt.readthedocs.io/en/stable/guide.html#trimming-paired-end-reads
-  cutadapt-se: " "
+  
+  cutadapt-se:
+    adapters: "-a AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC"
+    extra: "-q 20"
```
johanneskoester, line 131: Ensure that all changes in the config structure are reflected in workflow/schemas/config.schema.yaml.

config/config.yaml
```diff
@@ -61,6 +61,7 @@ diffexp:
     volcano-plot: 0.05
     ma-plot: 0.05
     qq-plot: 0.05
+  pre-define-genelist: "resources/genelist.tsv"
```
johanneskoester, line 64:
```suggestion
  # Optional (comment in to use): provide a list of genes that shall be shown in a heatmap
  # and for which bootstrap plots (see below) shall be created.
  # pre-define-genelist: "resources/genelist.tsv"
```

And below, in the bootstrap_plots section, remove the genes_of_interest key and adapt the code.

workflow/rules/trim_3prime.smk
```diff
@@ -0,0 +1,88 @@
...
+    #   * Removes reads having polyA (18bp) or polyG (18bp) and repeat the process twice by -n 2
+    # In short, it removes reads that are predominantly polyA or polyG.
+
+    # Rule cutadapt2 checks if reads are obtained from nextseq and removes adapters specific to nextseq
+    # reasoning behind parameters:
+    #   * `-m 20`:
+    #   * Discards any read under 20bp by -m 20
+    #   * `--nextseq-trim=10`
+    #   * Quality score<20 is trimmed, based on NextSeq/NovaSeq 2-color SBS system
+    #   * `-a A{18}AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=3;max_error_rate=0.100000`
```
johanneskoester, line 17: In order to make it clear that the current implementation just works for this specific lexogene setup, we should add a key to the config:
```
experiment:
  3-prime-rna-seq:
    # Specify vendor of the used protocol. Currently, only lexogene is supported.
    vendor: lexogene
```
And check for this value in this section of the code, raising and error if the vendor is not defined as lexogene.

johanneskoester (changes requested):

workflow/scripts/spia.R
```diff
@@ -17,60 +17,76 @@ db <- readRDS(snakemake@input[["spia_db"]])
...
+        )
+        res_reorder <- res_reorder %>%
+            rename(
+                "pGFdr" = "Combined Bonferroni p-values",
+                "tA" = "total perturbation accumulation",
+                "pSize" = "number of genes on the pathway",
+                "NDE" = "number of DE genes per pathway",
+                "pG" = "Combined p-value",
+                "pPERT" = "p-value to observe a total accumulation",
+                "pNDE" = "p-value for at least NDE genes"
```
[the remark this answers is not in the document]
manuelphilip: As I understood from this example, https://sparkbyexamples.com/r-programming/dplyr-rename-column/. It works fine as stated above. I checked the pathway output file even though there are no significant terms, the table contains header information which is correct with how we assign here.

workflow/rules/diffexp.smk
```diff
@@ -264,15 +264,19 @@ rule plot_diffexp_heatmap:
 rule plot_heatmap_predefined:
     input:
         logcountmatrix_file="results/tables/logcount-matrix/{model}.logcount-matrix.tsv",
-        predef_genelist="resources/selected_gene_from_ref.txt",
+        predef_genelist=config["diffexp"]["pre-define-genelist"],
     output:
-        predefgene_png="results/heatmaps/predefgenes_heatmap.png",
+        predefgene_pdf=report(
```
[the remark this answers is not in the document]
manuelphilip: Noted. updated the code.

workflow/rules/trim_3prime.smk
manuelphilip: Noted. Updated the code. Just as a note the name of the vendor is `lexogen` instead of `lexogene`.

config/config.yaml
manuelphilip: Noted

workflow/rules/diffexp.smk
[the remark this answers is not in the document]
johanneskoester: But the output file is now still named differently here (predefgene_pdf). If you name it the same in both rules, the script can be even simpler, because the writing is the same. just the input to the heatmap function is generated differently.

workflow/rules/trim_3prime.smk
```diff
@@ -0,0 +1,88 @@
...
+
+    # Rule cutadapt1 checks and removes poly-A tails and sequence qualilty score <20.
+    # reasoning behind parameters:
+    #   * `-m 20`:
+    #   * Discards any read under 20bp by -m 20
+    #   * `-O 20 -a ""polyA=A{20}"" -a ""QUALITY=G{20}"" -n 2`:
+    #   * Removes reads having polyA (18bp) or polyG (18bp) and repeat the process twice by -n 2
+    # In short, it removes reads that are predominantly polyA or polyG.
+
+    # Rule cutadapt2 checks if reads are obtained from nextseq and removes adapters specific to nextseq
```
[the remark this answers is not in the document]
johanneskoester: This option applies a special trimming mode, but it always does this. It would also do it for non-nextseq reads. It is fine to keep the code like this (as we want the vendor key in the config which makes it clear), but the comment should not claim that we really check for nextseq reads here.

config/config.yaml
johanneskoester: And use it for both, the bootstrap plots and the heatmap.

workflow/rules/trim_3prime.smk
johanneskoester: This also has to be reflected in the schema.

config/config.yaml
johanneskoester: This also has to be reflected in the schema.

workflow/rules/diffexp.smk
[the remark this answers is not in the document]
manuelphilip: Now we have single rule and same output file for both context. Depending on the `predef_genelist` the script will produce the heatmap.

workflow/rules/trim_3prime.smk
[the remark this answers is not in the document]
manuelphilip: Noted. Updated the comment section

config/config.yaml
manuelphilip: Done.

workflow/rules/trim_3prime.smk
manuelphilip: Done.

.test/3-prime-config/config/config.yaml
```diff
@@ -7,8 +7,12 @@ experiment:
   3-prime-rna-seq:
     activate: true
     # this allows to plot QC of aligned read postion for specific transcripts (or 'all' transcripts)
+    # Specify vendor of the used protocol. Currently, only lexogene is supported.
```
johanneskoester, line 10:
```suggestion
    # Specify vendor of the used protocol. Currently, only lexogen is supported.
```

config/config.yaml
```diff
@@ -5,8 +5,10 @@ experiment:
   # If set to `true`, this option allows the workflow to analyse 3-prime RNA seq data obtained from Quantseq protocol by Lexogen.
   # For more information https://www.lexogen.com/quantseq-3mrna-sequencing/
   3-prime-rna-seq:
-    activate: false
+    activate: true
     # this allows to plot QC of aligned read postion for specific transcripts (or 'all' transcripts)
+    # Specify vendor of the used protocol. Currently, only lexogene is supported.
```
johanneskoester, line 10:
```suggestion
    # Specify vendor of the used protocol. Currently, only lexogen is supported.
```

workflow/rules/common.smk
```diff
@@ -51,6 +51,18 @@ is_3prime_experiment = (
     .get("3-prime-rna-seq", dict())
     .get("activate", False)
 )
+is_3prime_vendor = (
+    config.get("experiment", dict())
+    .get("3-prime-rna-seq", dict())
+    .get("vendor", dict())
```
johanneskoester, line 57:
```suggestion
    .get("vendor")
```

workflow/rules/common.smk
```diff
@@ -51,6 +51,18 @@ is_3prime_experiment = (
...
+)
+
+if is_3prime_experiment:
+    if not is_3prime_vendor == "lexogen":
```
johanneskoester, line 61:
```suggestion
    if 3prime_vendor != "lexogen":
```

workflow/rules/common.smk
```diff
@@ -51,6 +51,18 @@ is_3prime_experiment = (
     .get("3-prime-rna-seq", dict())
     .get("activate", False)
 )
+is_3prime_vendor = (
```
johanneskoester, line 54:
```suggestion
3prime_vendor = (
```

workflow/scripts/plot_diffexp_heatmap.R
```diff
@@ -10,9 +10,12 @@ library(tidyr)
 sleuth_file <- read.csv(snakemake@input[["logcountmatrix_file"]],
     sep = "\t", header = TRUE
 )
+print(length(snakemake@params[["predef_genelist"]]))
+print(snakemake@params[["predef_genelist"]]$activate)
 # Getting pre-defined genelist file
-if (length(snakemake@input[["predef_genelist"]]) != 0) {
-    predefine_genelist <- read.table(snakemake@input[["predef_genelist"]],
+if (snakemake@params[["predef_genelist"]]$activate == TRUE &&
+    length(snakemake@params[["is_3prime_experiment"]]) == "true") {
```
johanneskoester, line 17: Why do you check for 3prime at all here? A heatmap for a predefined list of genes makes sense in any case.

workflow/scripts/plot_diffexp_heatmap.R
```diff
@@ -26,29 +29,76 @@ if (length(snakemake@input[["predef_genelist"]]) != 0) {
...
+        txt <- "cannot plot, all values are zero"
+        pdf(snakemake@output[["diffexp_heatmap"]], height = 10, width = 10)
+        plot.new()
+        text(.5, .5, txt, font = 2, cex = 1.5)
+        dev.off()
+    } else {
+        pdf(snakemake@output[["diffexp_heatmap"]], height = 10, width = 10)
+        pheatmap(sleuth_file[selectedgenes, ], scale = "row")
+        dev.off()
+    }
```
johanneskoester, line 103: This is still extremely repetitive. Basically it is always the same. pheatmap called on a dataframe of selected genes (either by significance or gene list), and generating an empty plot if that dataframe happens to be empty.

johanneskoester (changes requested):

workflow/scripts/plot_diffexp_heatmap.R
manuelphilip: Yes. Good catch!. I didn't think about it at all. Thank you for pointing it out. I too thought the same and since we have both RNA-seq and 3-prime, I always think in those two aspects. I have updated the code and removed unnecessary repetition.

workflow/scripts/plot_diffexp_heatmap.R
manuelphilip: Noted. As stated below I removed the 3prime part. It now checks only if the predefined list is there or not.

workflow/report/plot-heatmap.rst
johanneskoester, line 2:
```suggestion
Heatmap of top differentially expressed transcripts using the model ``{{ snakemake.params.model["full"] }}``.
```

workflow/rules/diffexp.smk
```diff
@@ -262,6 +242,26 @@ rule logcount_matrix:
...
 
+rule plot_diffexp_heatmap:
+    input:
+        logcountmatrix_file="results/tables/logcount-matrix/{model}.logcount-matrix.tsv",
+    output:
+        diffexp_heatmap=report(
+            "results/plots/diffexp-heatmap/{model}.diffexp-heatmap.pdf",
+            caption="../report/plot-heatmap.rst",
+            category="Heatmaps",
+        ),
```
johanneskoester, line 253: This rule does not yet behave as I would have expected.
I would do it in the following way:
```suggestion
    output:
        diffexp_heatmap=report(
            "results/plots/diffexp-heatmap/{model}.diffexp-heatmap.{mode}.pdf",
            caption="../report/plot-heatmap.rst",
            category="Heatmaps",
        ),
    wildcard_constraints:
        mode="topn|predefined"
```
We have a wildcard for the mode (top n differentially expressed genes, or predefined list). Depending on the mode, the script should do the following:

* `snakemake@wildcards[["mode"]] == "topn"`: fill `selectedgenes` such that it selects the top 50 genes from the differential expression dataframe (sorted by p-value).
* `snakemake@wildcards[["mode"]] == "predefined": fill `selectedgenes` such that it selects the genes from the predefined list.

This is an if-else at the beginning. The rest of the script should be outside of that if-else, and just generate the heatmap based on the selectedgenes variable.

workflow/rules/diffexp.smk
```diff
@@ -262,6 +242,26 @@ rule logcount_matrix:
...
+    params:
+        model=get_model,
+        predef_genelist=config["diffexp"]["genes_of_interest"],
```
johanneskoester, line 256: This has to be an input file, not a param. Also, remember that this can be missing in the config. Hence you need to put it into a input function which returns `[]` if no genes_of_interest are defined in the config.

johanneskoester (changes requested):

workflow/rules/diffexp.smk
manuelphilip: Here, Since this is the final output, We need to define it in common.smk. In that case, I am a bit confused regarding the wildcard_constraints. The mode has to be dependent upon the input file `predef_genelist`. So whether the mode has to be defined in the function which checks if the file exists should return mode as `predefined` or else ` topn`?

johanneskoester: If the predefined genelist is given in the config, I want both heatmaps, otherwise, I only want the topn.

johanneskoester: So this is what you request then in the final output in common smk. And in the first case, the rule is executed twice (once with mode predefined, once with topn), in the other case, you request only topn in the final output, such that the rule is only run once.

manuelphilip: The rule now runs two times based on the predefined genelist set to `true` and once if it is set to `false` given in the config.yaml file.

workflow/rules/diffexp.smk
manuelphilip: Noted. Added the function in `common.smk` file.

workflow/rules/common.smk
```diff
@@ -255,14 +262,34 @@ def all_input(wildcards):
...
+        wanted_input.extend(
+            expand(
+                [
+                    "results/plots/diffexp-heatmap/{model}.diffexp-heatmap.{mode}.pdf",
+                ],
+                model=config["diffexp"]["models"],
+                mode=["topn"],
+            )
+        )
+        return wanted_input
```
johanneskoester, line 292: Why this early return?

johanneskoester: This way, the targets below won't be collected.

workflow/scripts/plot_diffexp_heatmap.R
```diff
@@ -24,47 +42,31 @@ if (snakemake@params[["predef_genelist"]]$activate == TRUE) {
...
+    )
+    dev.off()
 } else {
-    # Adding gene name to corresponding transcript id from sleuth file
-    rownames(sleuth_file) <- paste(
-        sleuth_file$transcript,
-        ":", sleuth_file$gene
+    pdf(
+        file = snakemake@output[["diffexp_heatmap"]],
+        height = 10, width = 10
```
johanneskoester, line 68: Why having this hardcoded? Does the number of cols not play a role here like above with nrow > 100? Why does above case not always work?

johanneskoester (changes requested):

workflow/rules/common.smk
manuelphilip: Sorry, That was a typo mistake. Removed it.

workflow/scripts/plot_diffexp_heatmap.R
manuelphilip: As I understood the pdf function does not dynamically adjust the height and width. For the above case, we are selecting the top 50 variable genes, which by default remain the same (number of cols). The above case ( nrow > 100) also works here and I have updated the code.

.github/workflows/main.yml
```diff
@@ -42,13 +70,19 @@ jobs:
       with:
         submodules: recursive
 
-    - name: Test workflow
+#    - name: Test workflow
+#      uses: snakemake/snakemake-github-action@v1.23.0
+#      with:
+#        directory: .test
+#        snakefile: workflow/Snakefile
+#        args: "--use-conda --show-failed-logs --cores 2 --conda-cleanup-pkgs cache --all-temp"
```
johanneskoester, line 78: Why is this now commented out??

johanneskoester (changes requested):

.github/workflows/main.yml
manuelphilip: I tried to run both simultaneously, but when the workflow reached 3-prime part, it fails due to lack of data space. So I comment one after it is finished and ran the other.

config/config.yaml
```diff
@@ -52,9 +65,14 @@ diffexp:
     volcano-plot: 0.05
     ma-plot: 0.05
     qq-plot: 0.05
+  # Optional (comment in to use): provide a list of genes that shall be shown in a heatmap
+  # and for which bootstrap plots (see below) shall be created.  
+  genes_of_interest:
+    activate: false
+    genelist: "resources/gene_list.tsv"
```
johanneskoester, line 72:
```suggestion
  # genes_of_interest:
  #   activate: false
  #   genelist: "resources/gene_list.tsv"
```

config/config.yaml
```diff
@@ -40,7 +52,8 @@ diffexp:
   models:
     model_X:
       full: ~condition + batch_effect
-      reduced: ~batch_effect
+      reduced: batch_effect
+      #reduced: '~1'# in case if it takes as numerical
```
johanneskoester, line 56:
```suggestion
      reduced: ~batch_effect
```

config/config.yaml
```diff
@@ -52,9 +65,14 @@ diffexp:
...
     qq-plot: 0.05
+  # Optional (comment in to use): provide a list of genes that shall be shown in a heatmap
+  # and for which bootstrap plots (see below) shall be created.  
+  genes_of_interest:
+    activate: false
+    genelist: "resources/gene_list.tsv"
 
 diffsplice:
-  activate: true
+  activate: false
```
johanneskoester, line 75:
```suggestion
  activate: true
```

workflow/rules/common.smk
```diff
@@ -202,12 +262,33 @@ def all_input(wildcards):
...
+    else:
+        wanted_input.extend(
+            expand(
+                [
+                    "results/plots/diffexp-heatmap/{model}.diffexp-heatmap.{mode}.pdf",
+                ],
+                model=config["diffexp"]["models"],
+                mode=["topn"],
+            )
+        )
```
johanneskoester, line 291: Redundant, just change what you pass to mode depending on whether genes_of_interest is activated or not.

workflow/rules/trim_3prime.smk
```diff
@@ -0,0 +1,88 @@
+if is_3prime_experiment:
```
johanneskoester, line 1: and check whether vendor is lexogen!

johanneskoester (changes requested):

workflow/rules/common.smk
manuelphilip: A small doubt, As we have commented the `genes_of_interest` in the config file, I am not clear how can we pass `genes_of_interest` variable by default.

johanneskoester: You could use .get() here, but actually it is indeed better to not comment out genes_of_interest. Instead, just have activate: false as default for it, as you had before. I withdraw my change request for that.

config/config.yaml
johanneskoester: Not a good idea, I withdraw this.

johanneskoester: Can you please make genes_of_interest with its subitems a required element in the config schema?

manuelphilip: Noted. Added.

manuelphilip: Noted. Changed

config/config.yaml
manuelphilip: Noted. Changed

config/config.yaml
manuelphilip: Changed

workflow/rules/common.smk
manuelphilip: From what I understood based on the comments above, I updated the output based on the activation of `genes_of_interest` parameter.

workflow/rules/trim_3prime.smk
manuelphilip: Noted

workflow/rules/trim_3prime.smk
```diff
@@ -1,4 +1,4 @@
-if is_3prime_experiment:
+if is_3prime_experiment and three_prime_vendor:
```
johanneskoester, line 1: Must be three_prime_vendor == "lexogen":

johanneskoester (changes requested):

workflow/rules/trim_3prime.smk
manuelphilip: Noted. Added

workflow/rules/diffexp.smk
```diff
@@ -57,21 +57,25 @@ rule sleuth_diffexp:
             "results/plots/mean-var/{model}.mean-variance-plot.pdf",
             caption="../report/plot-mean-var.rst",
             category="QC",
+            labels={"model": "{model}-mean-variance-plot"},
```
johanneskoester, line 60: This is not a good label. Rather have multiple keys: `{"model": "{model}", "plot": "mean-variance"}`, analogously below.

workflow/rules/diffexp.smk
```diff
@@ -306,6 +308,7 @@ rule plot_fragment_length_dist:
             "results/plots/fld/{sample}-{unit}.fragment-length-dist.pdf",
             caption="../report/plot-fld.rst",
             category="Fragment length distribution",
+            labels={"samples": "{sample}-{unit}"},
```
johanneskoester, line 311: Put this into category `category="QC"`, `subcategory="per-sample"`, and then `labels={"sample": "{sample}-{unit}", "plot": "fragment lengths"}`. In addition, put all the the model-specific QC into `subcategory="per-model"`.

workflow/rules/qc_3prime.smk
```diff
@@ -58,6 +58,7 @@ if is_3prime_experiment and config["experiment"]["3-prime-rna-seq"]["plot-qc"] !
                 "results/plots/QC/3prime-ind-QC-plot.{ind_transcripts}.html",
                 category="QC",
                 caption="../report/plot-3prime-QC-histogram.rst",
+                labels={"QC-plot": "{ind_transcripts}-QC-plot"},
```
johanneskoester, line 61: This QC-plot naming is still super-unspecific, I mentioned that quite often now.
```
category="QC",
subcategory="global",
labels={"plot": "3-prime read positioning", "subset": "{ind_transcripts}"}
```

johanneskoester (changes requested):

workflow/rules/diffexp.smk
manuelphilip: Noted. Done

workflow/rules/qc_3prime.smk
manuelphilip: Noted. Done

johanneskoester (changes requested):

PR #67, opened 2023-02-05
fix: Feature/update cutadapt wrapper

Austin-s-h:
I encountered an error in this trimming rule due to 

```
  File "~/mambaforge/envs/snakemake/lib/python3.10/site-packages/snakemake/__init__.py", line 22, in <module>
    from snakemake.target_jobs import parse_target_jobs_cli_args
  File "~/mambaforge/envs/snakemake/lib/python3.10/site-packages/snakemake/target_jobs.py", line 4, in <module>
    from snakemake.common import parse_key_value_arg
  File "~/mambaforge/envs/snakemake/lib/python3.10/site-packages/snakemake/common/__init__.py", line 284, in <module>
    @contextlib.asynccontextmanager
AttributeError: module 'contextlib' has no attribute 'asynccontextmanager'
```

Tried running under Python 3.10 and 3.7, still encountered the same issue, but eventually tracked the rabbit hole down to this. 

Updating the wrapper for cutadapt slightly changes the syntax, but no modification to the config.yaml file is needed. Additionally, (optionally) I added multithreading, as cutadapt scales quite well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,4 +20,4 @@ install:
 script:
   - source activate snakemake
   # run the workflow
-  - snakemake --use-conda --directory .test || (for f in .test/logs/*; do echo $f; cat $f; done; for f in .test/logs/*/*; do echo $f; cat $f; done; exit 1)
+  - snakemake --cores all --use-conda --directory .test || (for f in .test/logs/*; do echo $f; cat $f; done; for f in .test/logs/*/*; do echo $f; cat $f; done; exit 1)
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -124,4 +124,4 @@ params:
   #     false positive adapter matches
   #   * `--minimum-overlap 7`: the cutadapt default minimum overlap of `5` did trimming on the level
   #     of expected adapter matches by chance
-  cutadapt-pe: "-j 0 -a ACGGATCGATCGATCGATCGAT -g GGATCGATCGATCGATCGAT -A ACGGATCGATCGATCGATCGAT -G GGATCGATCGATCGATCGAT --minimum-length 33 -e 0.005 --overlap 7"
+  cutadapt-pe: "-a ACGGATCGATCGATCGATCGAT -g GGATCGATCGATCGATCGAT -A ACGGATCGATCGATCGATCGAT -G GGATCGATCGATCGATCGAT --minimum-length 33 -e 0.005 --overlap 7"
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -113,7 +113,9 @@ params:
   # also, separate capitalised letter flags are required for adapters in
   # the reverse reads of paired end sequencing:
   # * https://cutadapt.readthedocs.io/en/stable/guide.html#trimming-paired-end-reads
-  cutadapt-se: ""
+  cutadapt-se:
+    adapters: "-a AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC"
+    extra: "-q 20"
   # reasoning behind parameters:
   #   * `--minimum-length 33`:
   #     * kallisto needs non-empty reads in current versions (fixed for future releases:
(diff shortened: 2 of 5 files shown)


config/config.yaml
```diff
@@ -124,4 +124,4 @@ params:
   #     false positive adapter matches
   #   * `--minimum-overlap 7`: the cutadapt default minimum overlap of `5` did trimming on the level
   #     of expected adapter matches by chance
-  cutadapt-pe: "-a ACGGATCGATCGATCGATCGAT -g GGATCGATCGATCGATCGAT -A ACGGATCGATCGATCGATCGAT -G GGATCGATCGATCGATCGAT --minimum-length 33 -e 0.005 --overlap 7"
+  cutadapt-pe: "-j 0 -a ACGGATCGATCGATCGATCGAT -g GGATCGATCGATCGATCGAT -A ACGGATCGATCGATCGATCGAT -G GGATCGATCGATCGATCGAT --minimum-length 33 -e 0.005 --overlap 7"
```
dlaehnemann, line 127: You should not have to specify `-j` or `--cores` explicitly, as the wrappers take care of passing on the `threads: ` specification. See the [respective wrapper code for the `pe` setup](https://github.com/snakemake/snakemake-wrappers/blob/4ec0c91f37e037479c40551463449d96830a548e/bio/cutadapt/pe/wrapper.py#L25) and the [respective wrapper code for the `se` setup](https://github.com/snakemake/snakemake-wrappers/blob/4ec0c91f37e037479c40551463449d96830a548e/bio/cutadapt/se/wrapper.py#L25).

Thus, I suggest to remove this change:
```suggestion
  cutadapt-pe: "-a ACGGATCGATCGATCGATCGAT -g GGATCGATCGATCGATCGAT -A ACGGATCGATCGATCGATCGAT -G GGATCGATCGATCGATCGAT --minimum-length 33 -e 0.005 --overlap 7"
```

workflow/rules/trim.smk
```diff
@@ -5,12 +5,13 @@ rule cutadapt_pe:
         fastq1="results/trimmed/{sample}-{unit}.1.fastq.gz",
         fastq2="results/trimmed/{sample}-{unit}.2.fastq.gz",
         qc="results/trimmed/{sample}-{unit}.qc.txt",
+    threads: 8
```
dlaehnemann, line 8: Thanks for adding this! Good point!

dlaehnemann (changes requested):
Many thanks for catching and debugging this! Just one little thing that I would switch back.

dlaehnemann:
Ah, I think it is failing because this string here is empty (and we have [a single end test case unit](https://github.com/sansterbioanalytics/rna-seq-kallisto-sleuth/blob/feature/update-cutadapt-wrapper/.test/config/units.tsv#L4)):
https://github.com/sansterbioanalytics/rna-seq-kallisto-sleuth/blob/6cc1c649c668a0d63ed33668f210d89c72db208f/.test/config/config.yaml#L107

Maybe just replicate the settings string from `cutadapt-pe` in the `.test/config/config.yaml` and simply omit the capital letter adapter specs (`-A` and `-G`, they are for the reverse reads in the paired end setup):
https://github.com/sansterbioanalytics/rna-seq-kallisto-sleuth/blob/6cc1c649c668a0d63ed33668f210d89c72db208f/.test/config/config.yaml#L118

Austin-s-h:
> Ah, I think it is failing because this string here is empty (and we have [a single end test case unit](https://github.com/sansterbioanalytics/rna-seq-kallisto-sleuth/blob/feature/update-cutadapt-wrapper/.test/config/units.tsv#L4)): https://github.com/sansterbioanalytics/rna-seq-kallisto-sleuth/blob/6cc1c649c668a0d63ed33668f210d89c72db208f/.test/config/config.yaml#L107
> 
> Maybe just replicate the settings string from `cutadapt-pe` in the `.test/config/config.yaml` and simply omit the capital letter adapter specs (`-A` and `-G`, they are for the reverse reads in the paired end setup): https://github.com/sansterbioanalytics/rna-seq-kallisto-sleuth/blob/6cc1c649c668a0d63ed33668f210d89c72db208f/.test/config/config.yaml#L118

Should have addressed this in my latest :)

dlaehnemann (approved):

.travis.yml
```diff
@@ -20,4 +20,4 @@ install:
 script:
   - source activate snakemake
   # run the workflow
-  - snakemake --use-conda --directory .test || (for f in .test/logs/*; do echo $f; cat $f; done; for f in .test/logs/*/*; do echo $f; cat $f; done; exit 1)
+  - snakemake --cores all --use-conda --directory .test || (for f in .test/logs/*; do echo $f; cat $f; done; for f in .test/logs/*/*; do echo $f; cat $f; done; exit 1)
```
dlaehnemann, line 23: BTW, I think this `.travis.yml` file is an artifact from a time, when CI tests were run via Travis CI. But we have long moved to GitHub Actions, so all testing harness definitions are found in `.github/workflows/`. I guess we can simply remove this file here in the near future.

PR #68, opened 2023-02-06
fix: Update sleuth conda environment and add multithreading to kallisto quant

Austin-s-h:
While encountering a few errors related to differential expression, I upgraded the environment for running Sleuth and increased the efficiency of the kallisto quant rule.

Also, found a bug in the samples schema, batch_effect was missing.

### DIFF
--- a/workflow/rules/quant.smk
+++ b/workflow/rules/quant.smk
@@ -1,20 +1,20 @@
 rule kallisto_index:
     input:
-        "resources/transcriptome.cdna.fasta",
+        fasta="resources/transcriptome.cdna.fasta",
     output:
-        "results/kallisto/transcripts.idx",
+        index="results/kallisto/transcripts.idx",
+    params:
+        extra="",  # optional parameters
     log:
         "results/logs/kallisto/index.log",
-    conda:
-        "../envs/kallisto.yaml"
-    shell:
-        "kallisto index -i {output} {input} 2> {log}"
+    wrapper:
+        "v1.23.1/bio/kallisto/index"
 
 
 rule kallisto_quant:
     input:
-        fq=get_trimmed,
-        idx="results/kallisto/transcripts.idx",
+        fastq=get_trimmed,
+        index="results/kallisto/transcripts.idx",
     output:
         directory("results/kallisto/{sample}-{unit}"),
     log:
@@ -28,6 +28,5 @@ rule kallisto_quant:
     threads: 8
     resources:
         mem_mb=10000,
-    shell:
-        "kallisto quant -i {input.idx} -o {output} "
-        "{params.extra} -t {threads} {input.fq} 2> {log}"
+    wrapper:
+        "v1.23.1/bio/kallisto/quant"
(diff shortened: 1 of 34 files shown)


workflow/schemas/samples.schema.yaml
```diff
@@ -8,6 +8,9 @@ properties:
   condition:
     type: string
     description: sample condition that will be compared during differential expression analysis (e.g. a treatment, a tissue time, a disease)
-
+  batch_effect:
+    type: string
+    description: The batch that the sample belongs to
+
```
dlaehnemann, line 14: As the column name used in the `config["diffexp"]["models"]["modelX"]` specifications of `full` and `reduced` models is arbitrary, so I'm not sure if we need to include this in the schema (even if we don't include among the `required:` fields listed below). Does specifying this here give you a clear benefit (like a more useful error message in some case) or was this just meant for consistency?

workflow/rules/quant.smk
```diff
@@ -23,6 +23,11 @@ rule kallisto_quant:
         extra=kallisto_params,
     conda:
         "../envs/kallisto.yaml"
+    # around 4 gb of memory usage with 1 thread. (hg38)
+    # over 8gb peak memory usage with 8 threads. (hg38)
```
dlaehnemann, line 27: Thanks for checking this out and documenting the info!

workflow/rules/quant.smk
```diff
@@ -23,6 +23,11 @@ rule kallisto_quant:
...
+    threads: 8
+    resources:
+        mem_mb=10000,
     shell:
         "kallisto quant -i {input.idx} -o {output} "
-        "{params.extra} {input.fq} 2> {log}"
+        "{params.extra} -t {threads} {input.fq} 2> {log}"
```
dlaehnemann, line 33: While we are at it, I'd suggest moving to using the respective wrapper. This makes maintenance easier in the long run:
https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/kallisto/quant.html

Same for the `rule kallisto_index` above:
https://snakemake-wrappers.readthedocs.io/en/stable/wrappers/kallisto/index.html

dlaehnemann (changes requested):
Thanks again for taking the initiative of improving the workflow! I've added some suggestions.

workflow/schemas/samples.schema.yaml
Austin-s-h: It turned out not to be so arbitrary, I think that it should be enforced.

workflow/rules/quant.smk
Austin-s-h: Should be resolved! Please review @dlaehnemann

workflow/schemas/samples.schema.yaml
dlaehnemann: Ah, but it should be. If you simply name such a column differently in the `samples.tsv` file, for example `day_of_processing`, you can simply adjust the formulas in `config.yaml` to:
```
full: ~condition + day_of_processing
reduced: ~day_of_processing
```

Or if you don't have any batch effects, you can also do something like:
```
full: ~condition
reduced: ~1
```

Is this explanation clearer?

PR #73, opened 2023-06-02
fix: Updated the get-transcript-info.R file and its dependencies

manuelphilip:


workflow/scripts/get-transcript-info.R
```diff
@@ -39,69 +40,124 @@ while ( class(mart)[[1]] != "Mart" ) {
...
   attributes <- c(attributes, "transcript_is_canonical")
 }
 
-t2g <- biomaRt::getBM(
-            attributes = attributes,
-            mart = mart,
-            useCache = FALSE
-            )
 if (!has_canonical) {
   t2g <- t2g %>% add_column(transcript_is_canonical = NA)
```
johanneskoester, line 70: This will fail in case has_canonical is false, because t2g is not created above.

workflow/scripts/get-transcript-info.R
```diff
@@ -39,69 +40,124 @@ while ( class(mart)[[1]] != "Mart" ) {
...
-            )
-          }
+  attributes <- c(attributes, "transcript_is_canonical",
+    "chromosome_name", "transcript_mane_select",
+      "ensembl_transcript_id_version")
+  t2g <- biomaRt::getBM(
+  attributes = attributes,
+  mart = mart,
+  useCache = FALSE
+  )
```
johanneskoester, line 81: Please use indentation in line with the R style guide.

workflow/scripts/get-transcript-info.R
```diff
@@ -39,69 +40,124 @@ while ( class(mart)[[1]] != "Mart" ) {
...
+                TRUE
+              } else if (v == "0") {
+                FALSE
+              }
+            }
+          )
+        )
+      }
+    )
+  }
```
johanneskoester, line 162: This code block looks quite redundant. Please avoid copy-pasting of code and use function calls instead. Further, please add comments explaining what you do and why you do it for each operation.

johanneskoester (changes requested):

workflow/scripts/get-transcript-info.R
manuelphilip: Fixed the case

workflow/scripts/get-transcript-info.R
manuelphilip: I follow the R style that I installed in visual studio code and updated the R style.

workflow/scripts/get-transcript-info.R
manuelphilip: Updated the code to reduce redundancy. Added comments explaining each operation.

workflow/scripts/get-transcript-info.R
```diff
@@ -39,69 +40,88 @@ while ( class(mart)[[1]] != "Mart" ) {
...
         )
-
+      )
+    }
+  )
+# Filter transcipts that are canonical, mane selected and filter chromosomes that are defined as "patch" 
+if (three_prime_activated && has_canonical) {
+  t2g <- t2g %>%
+    filter(!str_detect(chromosome_name, "patch|PATCH")) %>%
+    filter(str_detect(transcript_mane_select, ""))
```
johanneskoester, line 125: So, if we only use mane select here, why do we have the has_canonical stuff above? Shouldn't we rather check for mane_select? Also, doesn't three_prime_activated already imply that canonical/mane_select attributes are present? And if not, I think there should rather be an error saying that we need canonical/mane_select if three prime mode is activated.

workflow/scripts/get-transcript-info.R
```diff
@@ -39,69 +40,88 @@ while ( class(mart)[[1]] != "Mart" ) {
...
+              TRUE
+            } else if (v == "0") {
+              FALSE
+            }
           }
         )
-
+      )
+    }
+  )
```
johanneskoester, line 120: Shouldn't such a treatment be applied to mane_select as well?

johanneskoester (changes requested):
Very good, looks much cleaner now! Some further comments below.

workflow/scripts/get-transcript-info.R
manuelphilip: Noted. Yes and I have added and else if case if three_prime is activated and it mane_select is not present.

workflow/scripts/get-transcript-info.R
manuelphilip: We are not using `mane_select` column in any of the output tables as we use canonical. So I am not sure if we need to add this for mane_select.

workflow/scripts/get-transcript-info.R
```diff
@@ -39,69 +40,95 @@ while ( class(mart)[[1]] != "Mart" ) {
...
+if (three_prime_activated && has_mane_select) {
+  t2g <- t2g %>%
+    filter(!str_detect(chromosome_name, "patch|PATCH")) %>%
+    filter(str_detect(transcript_mane_select, ""))
+}else if (three_prime_activated && !has_mane_select) {
+  stop(
+    str_c(
+        "needed mane_selected column in biomart if three prime mode is activated"
+          ))
+}
```
johanneskoester, line 133: This can be written in a much simpler way. First check for only three_prime_activated, and in that if handle both cases about has_mane_select or not.

johanneskoester (changes requested):

workflow/scripts/get-transcript-info.R
manuelphilip: Noted. Thank you pointing it. Updated the code.

johanneskoester (approved):

PR #75, opened 2023-08-15
fix: update samtools.yaml to latest `1.17` and update github actions

dlaehnemann:
@manuelphilip Was there a reason for the old version of `samtools =1.9`? Otherwise I would update this to the latest version.

I found this, because we were seeing very large memory usage (`> 20GB`) for the `rule get_mapped_canonical_transcripts`, which should stream through the BAM (and the BAMs we had were below `1GB`), so I wouldn't expect any large memory usage. The only explanation I have so far, is that maybe this very old version of samtools somehow had a bug or very suboptimal handling of BAM files.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Formatting
-      uses: github/super-linter@v4
+      uses: github/super-linter@v5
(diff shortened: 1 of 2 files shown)


manuelphilip:
Hi @dlaehnemann , sure you can update it to latest version and it should run

johanneskoester (approved):

PR #78, opened 2023-08-17
fix: simpler three prime QuantSeq cutadapt setup

dlaehnemann:
I have tested this on actual QuantSeq data, and the cutadapt setup in this PR yields mostly the same results that the complicated previous three step setup recommended by Lexogen gave. It seems to be a bit more aggressive in the poly-A tail removal, but this should not affect results much. And it greatly simplifies the workflow setup.


workflow/rules/trim.smk
```diff
@@ -12,22 +12,33 @@ rule cutadapt_pe:
     log:
         "results/logs/cutadapt/{sample}-{unit}.log",
     wrapper:
-        "v1.22.0/bio/cutadapt/pe"
+        "v2.6.0/bio/cutadapt/pe"
 
 
-if not is_3prime_experiment:
+rule cutadapt:
```
manuelphilip, line 18: I have a doubt here, as mentioned in the `https://faqs.lexogen.com/faq/what-is-the-adapter-sequence-i-need-to-use-for-t-1` the reads are initially removed for polyA tails, then the adapters are removed in two separate Cutadapt steps. I am not sure if replacing it with a single command works as mentioned. Did you manage to check it with Quantseq reads?

manuelphilip (commented):
I have a doubt regarding the cutadapt step that I mentioned in the corresponding section.

config/README.md
```diff
@@ -1,16 +1,72 @@
...
+* `-a "QUALITY=G{20}"`: We replace this with `cutudapt`s dedicated option for the removal artifactual trailing `G`s in NextSeq and NovaSeq data: `--nextseq-trim=20`.
+* `-n`: With the dedicated `cutadapt` options getting applied in the right order, this option is not needed any more.
+* `-a "r1adapter=A{18}AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=3;max_error_rate=0.100000"`: We remove A{18}, as this is handled by `--poly-a`. We increase `min_overlap` to 7 and set the `max_error_rate` to the Illumina error rate of about 0.005, both to avoid spurious adapter matches being removed.
+* `-g "r1adapter=AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=20"`: This is not needed any more, as `-a` option will lead to complete removal of read sequence if adapter is found at the start of the read, see: https://cutadapt.readthedocs.io/en/stable/guide.html#rightmost
+* `--discard-trimmed`: We omit this, as the `-a` with the adapter sequence will lead to complete read sequence removal if adapter is found at start, and the `--minimum-length` will then discard such empty reads.
```
dlaehnemann, line 71: @manuelphilip This should explain all the details of the changes to the cutadapt command(s) from the original documentation that I made. As the cutadapt manual states, it will execute the respective trimming actions in the correct order, see here:
https://cutadapt.readthedocs.io/en/stable/reference.html#read-modification-order

Namely, it will do them in the following order:
1. Do quality trimming. `--nextseq-trim` is effectively a special version of quality trimming, so this should run first. And this makes the most sense, as the trailing `G`s that NextSeq data might have, will be at the very end of reads.
2. Trim off adapters specified with `-a`. These should be the outermost after quality trimming.
3. `--poly-a` tail removal should come after quality and adapter trimming, as the poly-A tails are part of the actual target sequence / fragments to be sequenced.

I double-checked with some minimal QuantSeq data from the original publication that you provided, but feel free to point any of your actual data analyses to this branch and try it out. Results are not exactly identical, but roughly identical. And I expect the new version provided here to actually be better, as both the `--nextseq-trim` and `--poly-a` flags are dedicated flags for the tasks that are supposed to be better than highjacking `-a` and `-g` for this, and they enforce the right order of trimming.

manuelphilip: Aha..This sound perfect. Thank you @dlaehnemann for investigating your time into cutadapt section. I will also check it with the inhouse reads, but I think it should give us similar results.

manuelphilip (approved):

dlaehnemann:
We have now tested this on some real QuantSeq data, and the results seem to add up, even though [there aren't any specific `--poly-a` statistics in the cutadapt log file](https://github.com/marcelm/cutadapt/issues/696#issuecomment-1607879589). But using `cutadapt 4.4 with Python 3.10.12` with these new recommended settings (using the standard cutadapt workflow rule):

```
--cores 8 -a r1adapter=AGATCGGAAGAGCACACGTCTGAACTCCAGTCAC;min_overlap=7;max_error_rate=0.005 --minimum-length 33 --nextseq-trim=20 --poly-a
```

We in this instance get:
```
=== Summary ===

Total reads processed:              11,901,529
Reads with adapters:                 1,680,520 (14.1%)

== Read fate breakdown ==
Reads that were too short:             461,240 (3.9%)
Reads written (passing filters):    11,440,289 (96.1%)

Total basepairs processed: 1,202,054,429 bp
Quality-trimmed:              10,535,736 bp (0.9%)
Total written (filtered):  1,035,645,758 bp (86.2%)
```

Even when accounting for the `adapter trimmed` basepairs recorded in the log, this leaves us with a gap of basepairs unaccounted for of:
```
Total basepairs processed:  1,202,054,429 bp
Quality-trimmed:          -    10,535,736 bp
adapter trimmed:          -    48,066,270 bp
Total written (filtered): - 1,035,645,758 bp (86.2%)
============================================
--poly-a + too short          107,806,665 bp (8,97%)
```

As noted, these have to be basepairs removed by removing reads that are too short, plus possibly the `--poly-a` option. So I guess these numbers are OK and we can go ahead and merge this PR.

PR #79, opened 2023-08-28
perf: bump datavzrd wrapper to 2.6.0 and general bug fixes

manuelphilip:


dlaehnemann (changes requested):
There are a lot of commits in this pull request that have changes that have long been introduced into the main branch. Could you try to identify the commits with the changes that are actually displayed when you look at the `Files changes` tab here on the pull request and then cherry-pick those onto a new branch that you start from the main branch? So something like:

```
git checkout main
git checkout -b datavzrd-2-6-0-and-plot-dimensions
git cherry-pick <commit-hash-1> <commit-hash-2> <commit-hash-3>
git push -u origin datavzrd-2-6-0-and-plot-dimensions
```

Then, the new PR will be much cleaner to merge. Let me know if I can help with anything along the way.

manuelphilip:
Hi David, I am confused about where and how to find `<commit-hash-1>` properly. Can you please assist me with the exact location to get this `<commit-hash-1>`?

dlaehnemann:
You can either find those right here on GitHub, by clicking on a particular commit. So I for example expect that the commit `perf bump datavzrd wrapper to 2.6.0 and general bug fixes` is one that belongs into this pull request. You can simply click on its link above, and then you should see the changes it introduces on a separate page. There, at the top right, you'll find `commit bd964f73ab69487921766290faf9674958ae1c58`. The letters and numbers are the commit's hash value.

Alternatively, you can find the commits locally on any machine where you have the repository cloned and the branch `3-prime-rna` checked out. You can then issue `git log` on the command line and will get a browsable list of all the commits. And each commit again start with an entry of something like `commit bd964f73ab69487921766290faf9674958ae1c58`.

So you just go through the list of commits here, and find all the hashes like `bd964f73ab69487921766290faf9674958ae1c58` for the commits that really introduce the changes of this pull request. And in the future, it is usually a good idea to start a new branch for any new pull request, especially if the original branch (like the `3-prime-rna` branch here) has already been merged. This keeps the commit history and pull requests much cleaner.

manuelphilip:
I understand it now. Thank you @dlaehnemann for the detailed explanation. It really helped. I have created a new pull request with the latest changes mentioned here. Shall I remove this pull request as this might create confusion with the new one that I created?

dlaehnemann:
With the new pull request, we can cross-reference this pull request here. So this should be pull request #80 , right?
With this mention, this old pull request over here gets mentioned in the new one, as well.

And as you probably don't need this branch here, any more, but will instead create new branches off of the `main` branch for any future changes, you could go and `Delete branch`. You will still have the branch locally, where you developed it, but the repository here will stay cleaner and saner.

manuelphilip:
Noted it.

PR #92, opened 2024-05-13
perf: datavzrd wrapper `v3.12.1`, offer-excel configurable, free disk space for CI, dynamic sleuth_init mem_mb, pure download rules as localrules

dlaehnemann:
this then uses `datavzrd=2.36.10`

### DIFF
--- a/.test/config/config.yaml
+++ b/.test/config/config.yaml
@@ -93,6 +93,9 @@ enrichment:
     # the species specified by resources -> ref -> species above
     pathway_database: "panther"
 
+report:
+  offer_excel: true
+
 bootstrap_plots:
   # desired false discovery rate for bootstrap plots, i.e. a lower FDR will result in fewer boxplots generated
   FDR: 0.01
--- a/.test/config/config.yaml
+++ b/.test/config/config.yaml
@@ -94,7 +94,7 @@ enrichment:
     pathway_database: "panther"
 
 report:
-  # make this 'true' or 'True', to get excel files for download in the snakemake
+  # make this `true`, to get excel files for download in the snakemake
   # report, BUT: this can drastically increase the runtime of datavzrd report
   # generation, especially on larger cohorts
   offer_excel: true
--- a/.test/config/config.yaml
+++ b/.test/config/config.yaml
@@ -61,7 +61,7 @@ diffexp:
     genelist: "resources/gene_list.tsv"
 
 diffsplice:
-  activate: true
+  activate: false
   # codingCutoff parameter of isoformSwitchAnalyzer, see
   # https://rdrr.io/bioc/IsoformSwitchAnalyzeR/man/analyzeCPAT.html
   coding_cutoff: 0.725
--- a/.test/config/config.yaml
+++ b/.test/config/config.yaml
@@ -94,6 +94,9 @@ enrichment:
     pathway_database: "panther"
 
 report:
+  # make this 'true' or 'True', to get excel files for download in the snakemake
+  # report, BUT: this can drastically increase the runtime of datavzrd report
+  # generation, especially on larger cohorts
   offer_excel: true
 
 bootstrap_plots:
--- a/workflow/resources/datavzrd/diffexp-template.yaml
+++ b/workflow/resources/datavzrd/diffexp-template.yaml
@@ -4,7 +4,7 @@ name: ?f"Differential expression analysis for model {wildcards.model}"
 datasets:
   genes_representative:
     path: ?input.genes_representative
-    offer-excel: true
+    offer-excel: ?params.offer_excel
     links:
       link to transcripts:
         column: ens_gene
@@ -21,7 +21,7 @@ datasets:
     separator: "\t"
   transcripts:
     path: ?input.transcripts
-    offer-excel: true
+    offer-excel: ?params.offer_excel
     links:
       link to genes representative:
         column: ens_gene
@@ -29,11 +29,11 @@ datasets:
     separator: "\t"  
   genes_aggregated:
     path: ?input.genes_aggregated
-    offer-excel: true
+    offer-excel: ?params.offer_excel
     separator: "\t"
   logcount_matrix:
     path: ?input.logcount_matrix
-    offer-excel: true
+    offer-excel: ?params.offer_excel
     links:  
       link to transcripts:
         column: transcript
--- a/workflow/resources/datavzrd/diffexp-template.yaml
+++ b/workflow/resources/datavzrd/diffexp-template.yaml
@@ -4,7 +4,7 @@ name: ?f"Differential expression analysis for model {wildcards.model}"
 datasets:
   genes_representative:
     path: ?input.genes_representative
-    offer-excel: ?params.offer_excel
+    offer-excel: ?params.offer_excel in ["true", "True"]
     links:
       link to transcripts:
         column: ens_gene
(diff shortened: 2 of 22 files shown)


.test/config/config.yaml
```diff
@@ -93,6 +93,9 @@ enrichment:
     # the species specified by resources -> ref -> species above
     pathway_database: "panther"
 
+report:
+  offer_excel: true
```
fxwiegand, line 97:
```suggestion
  offer_excel: false
```

Depending on the size of the input data this might cause extremely long runtimes as the excel writer in rust requires to read all of the input data into memory before one can write it out:

https://github.com/datavzrd/datavzrd/blob/556feb7b6a39745b42900d27fbcad26ee050e158/src/render/portable/mod.rs#L1944-L1981

dlaehnemann: Many thanks for checking the template and for the clarification!

Did you already check for better options for xlsx file writing? Have you for example seen this one:
https://docs.rs/simple-xlsx-writer/latest/simple_xlsx_writer/

At least it doesn't read everything into memory. But not sure, if it will be any quicker and if the interface works for how you parse stuff in datavzrd...

fxwiegand: This is literally the one we are using right now 😄

dlaehnemann: Hmmm. Maybe switch to something else, then... :thinking:

But yeah, I think this switch of the default here is good enough for now. I'll add a comment warning of the performance hit when activating this...

fxwiegand: > Hmmm. Maybe switch to something else, then... 🤔

I am not sure there is something that would allow something like flushing the writer every thousand records for example. But it'd great to have something like that for sure. Might even have to do with excel itself maybe?

> But yeah, I think this switch of the default here is good enough for now. I'll add a comment warning of the performance hit when activating this...

Yeah a warning when manually activated sounds good!

dlaehnemann:
For now, I'll turn off the reactivated testing of isoformSwitchAnalyzer, to get the tests to pass here and merge this PR.

And then, I'll fix isoformSwitchAnalyzer on a new branch, where we'll also fix the Ensembl cDNA fasta file, as the current failure is due to the Ensembl GTF and the Ensembl cDNA FASTA file being out of sync, see here:
https://www.r-bloggers.com/2020/01/comparing-ensembl-gtf-and-cdna/

Removing Ensembl cDNA FASTA redundancies (the same HGNC gene symbol appearing with multiple Ensembl `gene_id`s due to scaffolds and other issues), will probably remove most of the FASTA-only transcript IDs as well, so should be part of the isoformSwitchAnalyzer fix...

workflow/resources/datavzrd/diffexp-template.yaml
```diff
@@ -1,8 +1,10 @@
+__use_yte__: true
+
 name: ?f"Differential expression analysis for model {wildcards.model}"
 datasets:
   genes_representative:
     path: ?input.genes_representative
-    offer-excel: true
+    offer-excel: ?params.offer_excel in ["true", "True"]
```
johanneskoester, line 7: I don't like the check for true and True as strings. The yaml schema should ensure that the value in the config yaml is a boolean. And then, there is no need to have these additional checks here.

workflow/rules/common.smk
```diff
@@ -360,7 +356,7 @@ def all_input(wildcards):
         )
     )
 
-    if is_activated(config["diffsplice"]):
+    if config["diffsplice"]["activate"]:
```
johanneskoester, line 359: Why this change?

workflow/rules/datavzrd.smk
```diff
@@ -64,13 +14,21 @@ rule spia_datavzrd:
...
+        "v3.11.0/utils/datavzrd"
 
 
 rule diffexp_datavzrd:
     input:
-        config="results/datavzrd/diffexp/{model}.yaml",
+        # the config file may be a yte template, with access to input, params and wildcards
+        # analogous to Snakemake's generic template support:
+        # https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#template-rendering-integration
+        # For template processing, __use_yte__: true has to be stated in the config file
```
johanneskoester, line 29:
```suggestion
```

workflow/rules/datavzrd.smk
```diff
@@ -85,19 +43,22 @@ rule diffexp_datavzrd:
...
             labels={"model": "{model}"},
         ),
-    params:
-        model=get_model,
     log:
         "logs/datavzrd-report/diffexp.{model}/diffexp.{model}.log",
+    params:
+        extra="",
+        model=get_model,
+        offer_excel=lookup(within=config, dpath="report/offer_excel", default="false"),
```
johanneskoester, line 51: Default should be a boolean.

workflow/rules/datavzrd.smk
```diff
@@ -116,5 +77,8 @@ rule go_enrichment_datavzrd:
         ),
     log:
         "logs/datavzrd-report/go_enrichment-{model}/go_enrichment-{model}_{gene_fdr}.go_term_fdr_{go_term_fdr}.log",
+    params:
+        offer_excel=lookup(within=config, dpath="report/offer_excel", default="false"),
```
johanneskoester, line 81:
```suggestion
        offer_excel=lookup(within=config, dpath="report/offer_excel", default=False),
```

workflow/rules/common.smk
```diff
@@ -80,10 +80,6 @@ def check_config():
 check_config()
 
 
-def is_activated(config_element):
-    return config_element["activate"] in {"true", "True"}
-
-
 def get_model(wildcards):
```
johanneskoester, line 83: It was always not good that we allowed strings here. The value of activate has to be boolean and we should ensure that via the schema.

johanneskoester (changes requested):
Epic, thanks a lot! Just some minor things below.

fxwiegand:
It would probably also makes sense to upgrade to 2.36.15 as this will probably save loads of space in the resulting reports 😊

workflow/rules/common.smk
dlaehnemann: The `is_activated()` function was not working as intended (any more). Nowadays, the yaml seems to be parsing settings of `true` directly into a boolean, and that boolean was failing the string-based comparison. Also, this was the only place that function was still used. So instead of fixing the function, I removed it -- one layer less of abstraction.

workflow/resources/datavzrd/diffexp-template.yaml
dlaehnemann: Yeah, that's a good idea. Actually, I have no clue why the `true` or `false` in this case got parsed into a string, while in the diffsplice case it seems to have turned into a boolean. Neither of them was mentioned in the `config.schema.yaml` so far...

workflow/rules/datavzrd.smk
dlaehnemann: Already done in response to previous comment.

workflow/rules/datavzrd.smk
dlaehnemann: Already done in response to previous comment.

workflow/rules/common.smk
dlaehnemann: Already done in response to previous comment.

workflow/rules/common.smk
johanneskoester: ok

PR #94, opened 2024-05-29
feat: Allow bam input files

fxwiegand:
This PR allows bam files as input via the fields `bam_single` or `bam_paired` in `units.tsv`.

### DIFF
--- a/workflow/rules/bam.smk
+++ b/workflow/rules/bam.smk
@@ -1,0 +1,28 @@
+rule separate_bam_paired:
+  input:
+    "{bam_file}.bam"
+  output:
+    "{bam_file}.1.fq",
+    "{bam_file}.2.fq",
+  log:
+    "{bam_file}.separate.log",
+  params:
+    fastq="-n",
+  threads: 3
+  wrapper:
+      "v3.10.2/bio/samtools/fastq/separate"
+
+
+rule samtools_fastq_interleaved:
+  input:
+    "{bam_file}.bam",
+  output:
+    "{bam_file}.fq",
+  log:
+    "{bam_file}.interleaved.log",
+  params:
+    " ",
+  threads: 3
+  wrapper:
+    "v3.10.2/bio/samtools/fastq/interleaved"
+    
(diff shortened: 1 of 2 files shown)


workflow/rules/bam.smk
```diff
@@ -0,0 +1,27 @@
+rule separate_bam_paired:
```
johanneskoester, line 1:
```suggestion
rule bam_paired_to_fastq:
```

workflow/rules/bam.smk
```diff
@@ -0,0 +1,27 @@
...
+    log:
+        "{bam_file}.separate.log",
+    params:
+        fastq="-n",
+    threads: 3
+    wrapper:
+        "v3.10.2/bio/samtools/fastq/separate"
+
+
+rule samtools_fastq_interleaved:
```
johanneskoester, line 16:
```suggestion
rule bam_single_to_fastq:
```

workflow/rules/bam.smk
```diff
@@ -0,0 +1,27 @@
+rule separate_bam_paired:
+    input:
+        "{bam_file}.bam",
+    output:
+        "{bam_file}.1.fq",
+        "{bam_file}.2.fq",
```
johanneskoester, line 6: For this and the other rule below, I would suggest to not write the fastqs into the same dir as the bam. Instead, write to a fixed path under results with sample and unit as wildcards. Also, store fq.gz (samtools can do this compression itself).

johanneskoester (changes requested):

workflow/rules/common.smk
```diff
@@ -92,13 +92,20 @@ def get_model(wildcards):
 
 def is_single_end(sample, unit):
     """Determine whether unit is single-end."""
+    bam2_not_present = pd.isnull(units.loc[(sample, unit), "bam_paired"])
```
johanneskoester, line 95: rather bam_paired_not_present?

johanneskoester (changes requested):

johanneskoester (approved):

PR #99, opened 2024-06-24
docs: improve configs and configuration README.md

dlaehnemann:


fxwiegand (approved):
Thanks for the update! This makes the kallisto section much clearer.

PR #100, opened 2024-06-24
perf: Update samtools fast separate wrapper

fxwiegand:
Just a quick PR to update the samtools fast separate wrapper


dlaehnemann (approved):
LGTM, simply moves to the latest wrapper for a very recently added functionality. Shouldn't break / rerun anybody else's analysis.

I took the liberty to also update the accompanying single-end wrapper, to keep them in sync and void further separate update PRs.

PR #102, opened 2024-06-25
fix: Remove non-existent outputs in spia rule

fxwiegand:
A quick PR that removes some non-existent output tsvs causing an error inn the spia R script.


dlaehnemann (approved):
Thanks for the cleanup!

PR #103, opened 2024-06-25
fix: Fix missing output in spia.R when no significant genes are found

fxwiegand:


dlaehnemann (approved):
LGTM!

Thanks for again fixing an uncaught omission!

PR #104, opened 2024-07-17
docs: Fix units configuration docs for bam files

fxwiegand:
Just a quick PR to update the docs for the recent changes from #94.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -21,6 +21,8 @@ For each unit, provide either of the following:
 * The path to two pairead-end FASTQ files in the columns `fq1`, `fq2`.
 * The path to a single-end FASTQ file in the column `fq1`.
   For single-end data, you also need to specify `fragment_len_mean` and `fragment_len_sd`, which should usually be available from your sequencing provider.
+* The path to a single-end BAM file in the column `bam_single`
+* The path to a paired-end bam BAM file in the column `bam_paired`
 
 Missing values can be specified by empty columns or by writing `NA`.
 


dlaehnemann (approved):
LGTM. One thing we might think about, is making the `units.tsv` also work with just one of those options' columns present. That way, we wouldn't have to always drag around a bunch of empty columns. But this might entail changing quite a bit of code whenever these columns are called. What do you think?

fxwiegand:
> LGTM. One thing we might think about, is making the `units.tsv` also work with just one of those options' columns present. That way, we wouldn't have to always drag around a bunch of empty columns. But this might entail changing quite a bit of code whenever these columns are called. What do you think?

I absolutely see the point you're making. We would need to make the pandas calls on these lines a more robust so they can handle when the bam columns are not present: 

https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/8d9773e952323eaddcdbad55ee70a076c0203de2/workflow/rules/common.smk#L93-L112

Besides that these `lookup` calls from snakemake would cause trouble. I am not sure how they handle missing columns.

https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/8d9773e952323eaddcdbad55ee70a076c0203de2/workflow/rules/bam.smk#L6

https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/8d9773e952323eaddcdbad55ee70a076c0203de2/workflow/rules/bam.smk#L25

dlaehnemann:
You can nowadays offer a `default=""` argument in the `lookup()` function, to capture those: https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#the-lookup-function. We just have to enforce the latest Snakemake version, then. Would be willing to create a new PR referencing this discussion? Then we can continue discussing there, as well.

fxwiegand:
> You can nowadays offer a `default=""` argument in the `lookup()` function, to capture those: https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#the-lookup-function. We just have to enforce the latest Snakemake version, then. Would be willing to create a new PR referencing this discussion? Then we can continue discussing there, as well.

Yeah i will give it shot and we will discuss further in the new PR! This would also be quite nice for saving the versioning because now i think about it #94 was indeed a breaking change since the config **needs** the new bam columns for the workflow to run.

dlaehnemann:
Yeah, I think you are right. But I have honestly not payed enough attention to what's a breaking change in these workflows in the past. But we should. This regularly trips up users (us included). :sweat_smile:

fxwiegand:
> Yeah, I think you are right. But I have honestly not payed enough attention to what's a breaking change in these workflows in the past. But we should. This regularly trips up users (us included). 😅

Yep indeed - @Addimator #105 will let you remove the bam columns from your config again.

PR #105, opened 2024-07-17
fix: Handle missing bam columns in units.tsv

fxwiegand:
This PR makes the workflow handle missing bam columns introduced in #94 without throwing an error.

### DIFF
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -88,7 +88,7 @@ def get_model(wildcards):
 
 def is_single_end(sample, unit):
     """Determine whether unit is single-end."""
-    if 'bam_paired' in units.columns:
+    if "bam_paired" in units.columns:
         bam_paired_not_present = pd.isnull(units.loc[(sample, unit), "bam_paired"])
     else:
         bam_paired_not_present = True
@@ -99,9 +99,13 @@ def is_single_end(sample, unit):
 
 def get_fastqs(wildcards):
     """Get raw FASTQ files from unit sheet."""
-    if 'bam_single' in units.columns and not pd.isnull(units.loc[(wildcards.sample, wildcards.unit), "bam_single"]):
+    if "bam_single" in units.columns and not pd.isnull(
+        units.loc[(wildcards.sample, wildcards.unit), "bam_single"]
+    ):
         return f"results/fastq/{wildcards.sample}-{wildcards.unit}.fq.gz"
-    elif 'bam_paired' in units.columns and not pd.isnull(units.loc[(wildcards.sample, wildcards.unit), "bam_paired"]):
+    elif "bam_paired" in units.columns and not pd.isnull(
+        units.loc[(wildcards.sample, wildcards.unit), "bam_paired"]
+    ):
         fqfrombam1 = f"results/fastq/{wildcards.sample}-{wildcards.unit}.1.fq.gz"
         fqfrombam2 = f"results/fastq/{wildcards.sample}-{wildcards.unit}.2.fq.gz"
         return [fqfrombam1, fqfrombam2]
--- a/workflow/rules/common.smk
+++ b/workflow/rules/common.smk
@@ -102,9 +102,13 @@ def is_single_end(sample, unit):
 
 def get_fastqs(wildcards):
     """Get raw FASTQ files from unit sheet."""
-    if not column_missing_or_empty("bam_single", units, wildcards.sample, wildcards.unit):
+    if not column_missing_or_empty(
+        "bam_single", units, wildcards.sample, wildcards.unit
+    ):
         return f"results/fastq/{wildcards.sample}-{wildcards.unit}.fq.gz"
-    elif not column_missing_or_empty("bam_paired", units, wildcards.sample, wildcards.unit):
+    elif not column_missing_or_empty(
+        "bam_paired", units, wildcards.sample, wildcards.unit
+    ):
         return expand(
             "results/fastq/{sample}-{unit}.{read}.fq.gz",
             sample=wildcards.sample,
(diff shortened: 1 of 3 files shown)


fxwiegand:
@dlaehnemann Just as we discussed in #104.

I removed the columns from the test config so the CI should show us if the workflow handles the missing columns. I dont even thing we need to set a default value in the `lookup()` functions as these rules will only be executed if the columns are present in the units sheet.

workflow/rules/common.smk
```diff
@@ -88,16 +88,23 @@ def get_model(wildcards):
...
+        units.loc[(wildcards.sample, wildcards.unit), "bam_single"]
+    ):
         return f"results/fastq/{wildcards.sample}-{wildcards.unit}.fq.gz"
-    elif not pd.isnull(units.loc[(wildcards.sample, wildcards.unit), "bam_paired"]):
+    elif "bam_paired" in units.columns and not pd.isnull(
+        units.loc[(wildcards.sample, wildcards.unit), "bam_paired"]
+    ):
         fqfrombam1 = f"results/fastq/{wildcards.sample}-{wildcards.unit}.1.fq.gz"
         fqfrombam2 = f"results/fastq/{wildcards.sample}-{wildcards.unit}.2.fq.gz"
         return [fqfrombam1, fqfrombam2]
```
dlaehnemann, line 110: A slight cosmetic / idiomatic suggestion:
```suggestion
        return expand("results/fastq/{sample}-{unit}.{read}.fq.gz",
            sample = wildcards.sample,
            unit = wildcards.unit,
            read = ["1", "2"],
        )
```

fxwiegand: Sure look good - will commit when the current CI run has finished as i do not want to interrupt the tests.

dlaehnemann:
I just realised, that we also do not have a test case for the new bam input functionality, and we probably should. Otherwise things are sure to break at some point. I see two possibilities, here:

1. Extend what we are doing in the QuantSeq tests right here, by introducing a rule that maps one pair of files and then we use the resulting bam file in our units.tsv as another input.
2. Include some bam file by mapping the existing fqs in the separate ngs-test-data repo and then write up a test case based on that?

Or maybe you know a (very minimal) bam file that we can directly use, optimally in some stable repository for download.

And it probably makes sense to include these tests in this PR, as this is the functionality that it should test...

fxwiegand:
I agree that a test case with the bam input would be helpful although really all it will test is the modified fastq input function as well as the samtools separate and interleaved wrapper (that should already be tested well enough).

I would argue to merge and release this now and add a test for the bam input later as this will at least allow any users to use the newest workflow version with the regular fastq inputs without any breaking changes.

workflow/rules/common.smk
```diff
@@ -88,19 +88,29 @@ def get_model(wildcards):
 
 def is_single_end(sample, unit):
     """Determine whether unit is single-end."""
-    bam_paired_not_present = pd.isnull(units.loc[(sample, unit), "bam_paired"])
+    if "bam_paired" in units.columns:
+        bam_paired_not_present = pd.isnull(units.loc[(sample, unit), "bam_paired"])
+    else:
+        bam_paired_not_present = True
     fq2_not_present = pd.isnull(units.loc[(sample, unit), "fq2"])
     return fq2_not_present and bam_paired_not_present
```
dlaehnemann, line 96: This might be over-engineering a bit, but we could make the code changes here a little less repetitive by creating the following function:

```
def column_missing_or_empty(column_name, dataframe, sample, unit):
    if column_name in dataframe.columns:
        return pd.isnull(dataframe.loc[(sample, unit), column_name])
    else:
        return True
```

Then the code here becomes:
```suggestion
    return column_missing_or_empty("fq2", units, sample, unit) and column_missing_or_empty("bam_paired", units, sample, unit)
```

workflow/rules/common.smk
```diff
@@ -88,19 +88,29 @@ def get_model(wildcards):
...
 
 
 def get_fastqs(wildcards):
     """Get raw FASTQ files from unit sheet."""
-    if not pd.isnull(units.loc[(wildcards.sample, wildcards.unit), "bam_single"]):
+    if "bam_single" in units.columns and not pd.isnull(
+        units.loc[(wildcards.sample, wildcards.unit), "bam_single"]
+    ):
```
dlaehnemann, line 103: With the function from above, this could become:
```suggestion
    if column_missing_or_empty("bam_single", units, wildcards.sample, wildcards.unit):
```

workflow/rules/common.smk
```diff
@@ -88,19 +88,29 @@ def get_model(wildcards):
...
         return f"results/fastq/{wildcards.sample}-{wildcards.unit}.fq.gz"
-    elif not pd.isnull(units.loc[(wildcards.sample, wildcards.unit), "bam_paired"]):
-        fqfrombam1 = f"results/fastq/{wildcards.sample}-{wildcards.unit}.1.fq.gz"
-        fqfrombam2 = f"results/fastq/{wildcards.sample}-{wildcards.unit}.2.fq.gz"
-        return [fqfrombam1, fqfrombam2]
+    elif "bam_paired" in units.columns and not pd.isnull(
+        units.loc[(wildcards.sample, wildcards.unit), "bam_paired"]
+    ):
```
dlaehnemann, line 107: And this could also use the newly defined function:
```suggestion
    elif column_missing_or_empty("bam_paired", units, wildcards.sample, wildcards.unit):
```

dlaehnemann (approved):
Good point for making this available quickly, as this un-breaks things. Let's think about a simple further test in a separate PR.

Also, the current version works fine, but I do have one suggestion to make it a bit less repetitive (and maybe a bit more intuitive to read). Feel free to use or not.

fxwiegand:
@dlaehnemann It seems like the suggestions caused some new bug 😬

dlaehnemann:
Oups. :see_no_evil:

dlaehnemann (approved):
To miss, or `not` to miss, that is the column.

PR #122, opened 2024-10-15
fix: Update max in memory rows for diffexp report

fxwiegand:
This PR increases the maximum in memory genes/transcripts for the diffexp datavzrd report.

## Summary by CodeRabbit

- **New Features**
	- Increased capacity for handling larger datasets in memory by adjusting the `max-in-memory-rows` parameter from 15,000 to 42,000.

- **Bug Fixes**
	- No bug fixes were included in this release.

- **Documentation**
	- Configuration file structure remains unchanged, ensuring clarity in dataset definitions and views.

### DIFF
--- a/workflow/resources/datavzrd/diffexp-template.yaml
+++ b/workflow/resources/datavzrd/diffexp-template.yaml
@@ -1,5 +1,5 @@
 __use_yte__: true
-max-in-memory-rows: 30000
+max-in-memory-rows: 42000
 
 name: ?f"Differential expression analysis for model {wildcards.model}"
 datasets:
--- a/workflow/resources/datavzrd/diffexp-template.yaml
+++ b/workflow/resources/datavzrd/diffexp-template.yaml
@@ -1,5 +1,5 @@
 __use_yte__: true
-max-in-memory-rows: 15000
+max-in-memory-rows: 30000
 
 name: ?f"Differential expression analysis for model {wildcards.model}"
 datasets:


workflow/resources/datavzrd/diffexp-template.yaml
```diff
@@ -1,5 +1,5 @@
 __use_yte__: true
-max-in-memory-rows: 15000
+max-in-memory-rows: 30000
```
dlaehnemann, line 2: I'm assuming you have tried this with a real dataset, right?

My main question would be, if this will usually cover the full table with a standard human transcriptome. It might at least be enough for `genes_representative` and `genes_aggregated`, depending on the total number of genes. Although I do get more gene identifiers from the Ensembl cdna file that we usually use as a reference:
```
❯ zgrep "^>" Homo_sapiens.GRCh38.cdna.all.fa.gz | grep -o -P "ENSG[^\.]+" | sort | uniq | wc -l
41116
```
Thus, if this does not hurt performance of datavzrd (or rather the resulting HTML table in a browser) too much, we could think about increasing this to `42000`, right?

dlaehnemann (commented):
Just a question / suggestion for the number used here...

workflow/resources/datavzrd/diffexp-template.yaml
dlaehnemann: Was that testing on mouse? Or on human?

If it was on human, I would stick with the 30000, otherwise we could maybe test with human data and increase it accordingly. And/or make this configurable via the `config.yaml`, explaining the performance tradeoff? But we should also always be careful not to add too many configuration options, to keep the `config.yaml` manageable...

dlaehnemann: OK, let's go with 42000 for now, and make this configurable if we run into trouble with any reports not properly loading.

dlaehnemann (approved):

PR #123, opened 2024-10-16
fix: Convert pval and qval to heatmap

fxwiegand:
This PR converts pval and qval in the transcripts view of the diffexp report from bar plots to heatmaps.

## Summary by CodeRabbit

- **New Features**
	- Enhanced visualization of statistical values with heatmap representations for `pval` and `qval` in both `transcripts` and `genes_aggregated` datasets.
	- Updated domain values for improved clarity in significance levels.

These changes provide a more detailed and visually informative way to interpret data significance.

### DIFF
--- a/workflow/resources/datavzrd/diffexp-template.yaml
+++ b/workflow/resources/datavzrd/diffexp-template.yaml
@@ -211,18 +211,28 @@ views:
               url: "http://www.ensembl.org/Homo_sapiens/Gene/Summary?g={target_id}"
         pval:
           plot:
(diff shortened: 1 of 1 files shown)


fxwiegand:
> LGTM. So if you have tested that this all works, feel free to merge. One thing I would double-check, is that values above 0.25 work with this domain definition. But you're the datavzrd expert, here... ;)

Works for sure. I think the default for `clamp` is true so color wise this will get clamped but that is ok I guess (if not even intended).

PR #136, opened 2025-02-06
feat: Enrichment and pathway scatter-plots

Addimator:
This PR adds new scatter plot views to the pathway and enrichment datavzrd reports:
- pathway_scatter for the pathway report
- enrichment_scatter for the enrichment report

Changes & Features
- Each view includes an additional plot displaying the top 50 pathways/GO-terms.
- Users can individually highlight specific pathways/GO-terms for better visualization.

## Summary by CodeRabbit

- **New Features**
	- Added an interactive scatter plot for GO term enrichment analysis that enables users to highlight and select GO terms, clearly displaying metrics on both axes.
	- Introduced a new scatter plot view for pathway impact analysis with multi-selection capabilities, providing clear visualization of pathway perturbation and significance.
	- Integrated these visualizations into existing reports for enhanced data exploration.
	- New dataset entry for `go_terms_scatter` added to the enrichment analysis configuration.
	- New view `pathways_scatter` introduced for SPIA pathway impact analysis.
	- Added new rules for generating scatter plots from enrichment and pathway data.
	- Enhanced scatter plot functionality with a new Python script for improved visualization.
	- Updated formatting for dependency entries to ensure consistency.

### DIFF
--- a/workflow/resources/datavzrd/go-enrichment-template.yaml
+++ b/workflow/resources/datavzrd/go-enrichment-template.yaml
@@ -136,8 +136,10 @@ views:
   go_terms_scatter:
     dataset: go_enrichment
     desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the effect of the GO term for the {wildcards.model} model as determined by goatools.
-      The effect is determined as the average of all the absolute fold changes (all fold changes transformed to be >1) across all genes in a GO term.
-      After averaging the fold changes, the result is transformed to the log2 scale for reporting (and for use in the pi values we also report).
+      The effect is determined as the average of all the absolute fold changes across all genes in a GO term.
+      Here, absolute means a log2 fold change above 0, or a fold change above 1.
+      Thus, for averaging, a fold change of 0.5 would become 2, a fold change of 0.25 would become 4.
+      After averaging the fold changes, the result is transformed back to the log2 scale for reporting (and for use in the pi values we also report).
       Each point corresponds to a GO term.
       Click on a GO term to highlight it in colour.
       Several GO terms can be selected at the same time by holding down the Shift key."


workflow/scripts/enrichment_pathway_scatter.py
```diff
@@ -0,0 +1,184 @@
...
+                ),
+                alt.value("circle"),
+            ),
+        )
+        .transform_filter(selector)
+    )
+    return legend
+
+
+color_scheme = [
```
dlaehnemann, line 108: Maybe add a comment, why you manually add a color scheme here and what the colors are?

dlaehnemann (commented):
This generally looks good, just a couple of questions and suggestions. And obviously, I didn't check any actual output charts, but I guess your end users have provided a review of that, already. ;)

workflow/envs/pystats.yaml
```diff
@@ -6,4 +6,4 @@ dependencies:
   - pyreadr =0.5
   - altair =5.3
   - pyarrow =16.1
-  - vl-convert-python=1.7
+  - vl-convert-python =1.7
```
dlaehnemann, line 9: Just for info: the syntax convention for specifying version numbers in conda `.yaml` files always has a space before the range specifiers. I couldn't find a good source for the motivation behind this, but here's an example with the `six` version specification:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run

workflow/resources/datavzrd/go-enrichment-template.yaml
```diff
@@ -133,3 +133,11 @@ views:
               }  
             spec-path: ?input.vega_waterfall
             vega-controls: true
+  go_terms_scatter:
+    dataset: go_enrichment
+    desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the effect of the GO term for the {wildcards.model} model.
```
dlaehnemann, line 138: I think we should mention the tool here, and what exactly the x-axis value represents. Do you know, what value `goatools` actually puts into the `study_items:<value>` field? I just put `??XYZ??` where I didn't know what to say, and looking into goatools code and docs didn't help me find out... Maybe you already dug into this?
```suggestion
    desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the effect of the GO term for the {wildcards.model} model as determined by goatools.
    The effect is determined as the sum of all the absolute ??XYZ?? values across all genes in a GO term.
```

workflow/resources/datavzrd/spia-template.yaml
```diff
@@ -119,3 +119,12 @@ views:
               }  
             spec-path: ?input.vega_waterfall
             vega-controls: true
+  pathways_scatter:
+    dataset: spia_table
+    desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the total pertubation accumulation effect of the pathway for the {wildcards.model} model.
```
dlaehnemann, line 124: I think mentioning the actual tool used here can be useful context:
```suggestion
    desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the total pertubation accumulation effect of the pathway that SPIA determined for the {wildcards.model} model. 
```

dlaehnemann (approved):
This looks good. I would just do minimal adjustments to the captions / desriptions of the new plots. But feel free to merge whichever way you want to adjust them.

workflow/resources/datavzrd/go-enrichment-template.yaml
```diff
@@ -133,3 +133,12 @@ views:
               }  
             spec-path: ?input.vega_waterfall
             vega-controls: true
+  go_terms_scatter:
+    dataset: go_enrichment
+    desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the effect of the GO term for the {wildcards.model} model as determined by goatools.
+    The effect is determined as the sum of all the absolute ??XYZ?? values across all genes in a GO term.
```
dlaehnemann, line 139: Looks like I messed up the indentation, here. Also, we still need to resolve what `??XYZ??` actually is. Do you have any idea how exactly this is defined and calculated in goatools?

dlaehnemann (changes requested):
Just realized, that the CodeRabbit suggestion to add `name` as an argument to the `plot()` function is correct, to actively pass the `name` value into the function.

workflow/resources/datavzrd/go-enrichment-template.yaml
Addimator: Oh, I completely oversaw your second sentence, sorry... Furthermore I have no knowledge of how goatools works, until know I just used its results for visualizing :D Will have a look.

Addimator:
> Just realized, that the CodeRabbit suggestion to add `name` as an argument to the `plot()` function is correct, to actively pass the `name` value into the function.

Isn't this outdated? I use the "identifier" now instead of "name" and I pass it as an argument?

workflow/resources/datavzrd/go-enrichment-template.yaml
dlaehnemann: What are those values in the `study_items` column, though? Do you have an example file kicking around that we could have a look at. Because the docs don't explain it and I also didn't find it in the code.

Also, it seems like you introduced the `effect` column calculation in the following PR, so maybe you remember something about the calculation or the `study_items` field from back then?
https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/pull/115

Addimator: Oh sorry, I was busy with another task so I did not investigate much.

The effect column is the sum of the effects in study items (As you have already seen):
https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/7836d67eba55bb5deaaca231734293d5fbe5a2f3/workflow/scripts/postprocess_go_enrichment.py#L63-L69

If I understand it correct, the study items are the expression levels of the different genes in the go term.
image

Addimator: Oh sorry, I was busy with another task so I did not investigate much.

The effect column is the sum of the effects in study items (As you have already seen):
https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/7836d67eba55bb5deaaca231734293d5fbe5a2f3/workflow/scripts/postprocess_go_enrichment.py#L63-L69

If I understand it correct, the study items are the b_ fold changes of the different genes in the go term.
image


image

dlaehnemann: Ah, yes. That seems to clear it up. The values under `study_items` are then simply the beta values that sleuth provides, which are similar to a log2 fold changes. I also found this in our code (where I was not looking previously, so thanks for your pointers!), so it seems we are actually hacking their format:
https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/d0e6e785d0b2518f11efcf3d18a44a53bb578035/workflow/scripts/goatools-go-enrichment-analysis.py#L173
So I think I know how to adjust the description, then.

But, I think we also need to adjust the calculation of the sum of effects in the `postprocess_go_enrichment.py` right here:
https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/blob/d0e6e785d0b2518f11efcf3d18a44a53bb578035/workflow/scripts/postprocess_go_enrichment.py#L66C52-L66C71

As values are log2-values, adding them amounts to multiplying the actual fold changes. So I think we need to take `2^<log2-value>` before adding everything up, then maybe also dividing by the number of genes to get an average effect size per gene in the GO term (to avoid overweighing large GO terms) and then again taking the log2 of the result. But maybe @johanneskoester can also double-check my thinking, here.

dlaehnemann, line 139: This assumes that we stick with the sum of fold changes:
```suggestion
    The effect is determined as the sum of all the absolute fold changes (on a log2 scale) across all genes in a GO term.
```
Alternatively, we could go for an average of the fold changes. I'll open a pull request for this effect calculcation, so we can discuss and decide there.

Addimator: Then this PR should be fine :)

dlaehnemann: This should be addressed by PR #141, and the respective adjustment of this description that I did in this commit:
https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/pull/136/commits/fda25ee840951bd934cc9aa4458315655bf68c31

dlaehnemann: I addressed this via PR #141

workflow/resources/datavzrd/go-enrichment-template.yaml
```diff
@@ -133,3 +133,12 @@ views:
               }  
             spec-path: ?input.vega_waterfall
             vega-controls: true
+  go_terms_scatter:
+    dataset: go_enrichment
+    desc: ?f"The y-axis represents the False Discovery Rate (FDR), while the x-axis shows the effect of the GO term for the {wildcards.model} model as determined by goatools.
+      The effect is determined as the average of all the absolute fold changes (all fold changes transformed to be >1) across all genes in a GO term (reported on a log2 scale).
```
dlaehnemann, line 139:
```suggestion
      The effect is determined as the average of all the absolute fold changes (all fold changes transformed to be >1) across all genes in a GO term.
      After averaging the fold changes, the result is transformed to the log2 scale for reporting (and for use in the pi values we also report).
```

dlaehnemann (approved):
Everything's cleared up from my side, so I'll finally merge this! :tada:

PR #142, opened 2025-03-10
fix: Add support for Drosophila melanogaster

DanielM812:
I noticed two errors that occured while working with RNA-Seq data from Drosophila melanogaster and tried to fix them:

[1](https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/commit/698a4790b51cb25827cfc4c7769efe08b13ace19). Gene names for the fruit fly can sometimes have a colon in them, when belonging to a gene family (e.g. His4:CG33869 and His4:CG31611). This leads to an error when selecting the correct name and value in postprocess_go_enrichment.py as they are also separated by a colon. The separation then results in three parts (His4, CG33869, 0.73842) instead of two parts.
[2](https://github.com/snakemake-workflows/rna-seq-kallisto-sleuth/commit/075315f266a2e117a3a5dd0f8f1b4c338ac7a450). For Drosophila melanogaster the "ensembl_transcript_id_version" attribute from the Ensembl mart does not exist. This attribute is part of the 3-prime attributes which (I assume?) are not needed for standard RNA-Seq analysis.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transcript processing to dynamically detect and use available identifier attributes, enhancing output stability.
  - Enhanced gene value extraction to accurately handle varied input formats, ensuring more reliable results.

### DIFF
--- a/workflow/scripts/get-transcript-info.R
+++ b/workflow/scripts/get-transcript-info.R
@@ -90,7 +90,7 @@ wanted_attributes <- c(
 
 
 three_prime_attributes <- c(
-  "ensembl_transcript_id_version",
+  use_if_available("ensembl_transcript_id_version", available_attributes),
   "chromosome_name",
   "transcript_length",
   "strand"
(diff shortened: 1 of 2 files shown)


dlaehnemann (approved):
Thanks for these generalizations. Good catches, and sorry for the inconvenience.

PR #144, opened 2025-04-03
fix: Remove model filtering of logcount matrix

fxwiegand:
This PR removes the filtering of the logcount matrices. This way they contain the same gene set across different defined models.

## Summary by CodeRabbit

- **New Features**
	- Updated dataset configuration so that transcript links are now optional.
	- Modified data export to include all log count entries without filtering based on transcript presence.
- **Bug Fixes**
	- Disabled the filtering of columns containing only NA values in the samples data frame.


workflow/scripts/postprocess_logcount.py
```diff
@@ -8,10 +8,5 @@
 logcount_matrix = pd.read_csv(snakemake.input['logcount'], sep='\t')
 diffexp = pd.read_csv(snakemake.input['diffexp'], sep='\t')
 
-# Filter logcount_matrix to only include rows where 'transcript' is in 'target_id' of genes_representative
-filtered_logcount_matrix = logcount_matrix[
-    logcount_matrix['transcript'].isin(diffexp['target_id'])
-]
-
 # Save the filtered dataframe to a new TSV file
-filtered_logcount_matrix.to_csv(snakemake.output[0], sep='\t', index=False)
```
dlaehnemann, line 17: With these changes, the whole script just doesn't do anything apart from reading and writing the same file, right? So could we just remove the whole rule and rewire the downstream rule to use this rule's input?

workflow/scripts/sleuth-init.R
```diff
@@ -45,10 +45,6 @@ samples_out <- if(!is.null(model[["full"]])) {
 # store design matrix
 write_rds(samples_out, file = snakemake@output[["designmatrix"]])
 
-# remove all columns which have only NA values
-samples <- samples %>%
-	    select_if(function(col) !all(is.na(col)))
-
```
dlaehnemann, line 51: Are we sure we can safely remove this without any side-effects? Would be great to have a comment here, explaining why we do this filtering... :sweat_smile:

dlaehnemann (commented):
Just had a look and I think we can do a more thorough cleanup, here?

workflow/scripts/postprocess_logcount.py
fxwiegand: Yeah we could indeed remove it entirely. That was just the quickest way for me to test this it hink.

workflow/scripts/sleuth-init.R
fxwiegand: For my run it worked fine. The goal with this entire PR was the fact that the log count matrix couldn't be compared across different models since they would miss rows.

dlaehnemann: Yeah, it probably was initially done to keep the reported table(s) clean. But with datavzrd this now restricts to explicitly named columns for the report, anyways. So i guess this should be alright, especially if all of the tests properly pass.

workflow/scripts/postprocess_logcount.py
dlaehnemann: If you do this update, make sure to pull first. I merged in the latest changes from the `main` branch.

fxwiegand: Just did! Please have a look again if this looks good to you.

dlaehnemann (approved):
Yeah, this looks good now!

PR #148, opened 2025-04-16
fix: better error message for column missing or empty

dlaehnemann:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error messages to display the exact conflicting values found, making it easier to identify issues when multiple values are present for a given sample, unit, and column.


dlaehnemann:
Turned out to be unrelated, a doubling of another sample set in the units.tsv.

PR #149, opened 2025-04-16
fix: use `verify_integrity=True` in pandas `set_index()` for samples and units to check for duplicates

dlaehnemann:
This does an early check of duplicates during samples.tsv and units.tsv reading, avoiding cryptic error messages downstream.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved data validation by ensuring that index columns in sample and unit tables are unique, preventing potential issues caused by duplicate entries.


dlaehnemann:
This might be something we also want to do in other workflows.

